PR #1, opened 2016-02-16
Add matching capability to Rule instances

d00rman:
When defining a rule, the user may from now on specify a match: stanza which contains the match specification of the positive criteria to include when filtering received event messages. The match object is parsed and used in testing the incoming messages, which enter the processing pipeline iff all of the match components specified by the user match. The algorithm supports matching simple values (numbers, strings, booleans, etc.) as well as nested objects and regular expressions.

Moreover, Rule instances are also able to expand the match object using the current event message's context to give the concrete values matched in it (e.g. concrete regex results).

Example rule definition:

```yaml
topic: resource_changed
match:
  meta:
    uri: /html\/([^\/]+)(?:|\/([^\/]+))$/
  exec:
    - uri: http://localhost:7231/{message.meta.domain}/v1/page/html/{match.meta.uri[1]}/{match.meta.uri[2]}
      headers:
        cache-control: no-cache
```

Thanks to the expansion feature, one can reuse the match object and populate the request template with it.

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -8,10 +8,153 @@ var Template = HyperSwitch.Template;
 function Rule(name, spec) {
 
     this.name = name;
-    this.spec = spec;
-    this._processRule(); 
+    this.spec = spec || {};
+    this.exec = [];
+    this.topic = '';
+    this.match = {
+        all: true,
+        test: '',
+        obj: {}
+    };
+    this.noop = false;
+
+    this._processRule();
 
 }
+
+
+/**
+ * Tests the message against the compiled evaluation test
+ *
+ * @param {Object} message the message to test
+ * @return true if no match is set for this rule or if the message matches
+ */
+Rule.prototype.test = function(message) {
+
+    return this.match.all || this._match_test(message);
+
+};
+
+
+/**
+ * Expands the rule's match object with the given message's content
+ *
+ * @param {Object} message the message to use in the expansion
+ * @return {Object} the object containing the expanded match portion of the rule
+ */
+Rule.prototype.expand = function(message) {
+
+    return this._match_expand(message);
+
+};
+
+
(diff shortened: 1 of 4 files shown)


d00rman:
cc @wikimedia/services

lib/rule.js
```diff
@@ -8,18 +8,160 @@ var Template = HyperSwitch.Template;
...
+ * @return {Object} the object containing the expanded match portion of the rule
+ */
+Rule.prototype.expand = function(message) {
+
+    return this._match_expand(message);
+
+};
+
+
+Rule.prototype._match_test = function(message) {
```
Pchelolo: What is the point in having those 2 empty functions in the prototype?

d00rman: `Rule.test()` and `Rule.expand()` use them, and as the comments say, these are redefined only for rules that specify match criteria. Besides, a couple of wasted lines buy us readability.

lib/rule.js
```diff
@@ -8,18 +8,160 @@ var Template = HyperSwitch.Template;
...
-    this.spec = spec;
-    this._processRule(); 
+    this.spec = spec || {};
+    this.exec = [];
+    this.topic = '';
+    this.match = {
+        all: true,
+        test: '',
+        obj: {}
+    };
```
Pchelolo: Why are you storing the `test` and `obj` in the object properties? As I see they're only used on compilation, so converting them to local variables will decrease memory footprint.

d00rman: Good point. I think I was using them originally for tests, but don't any more afaik.

d00rman:
@Pchelolo updated

Pchelolo:
@d00rman :+1: I think it's ready to go

d00rman:
Travis is still on holidays ...

PR #3, opened 2016-04-14
Refactoring

Pchelolo:
1. Separated ugly kafka client promisification to a separate class, that will work as a proxy to kafka and encapsulate kafka configuration.
2. Separated rule management from rule execution and rule subscription. 

cc @wikimedia/services @d00rman

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -56,8 +56,8 @@ class Rule {
 
         // this is an object, we need to split it into components
         const subObj = fieldName ? {} : result;
-        const test = Object.keys(obj).map((key) =>
-            this._compileMatch(obj[key], subObj, `${name}['${key}']`, key))
+        const test = Object.keys(obj).map(
+            (key) => this._compileMatch(obj[key], subObj, `${name}['${key}']`, key))
         .join(' && ');
         if (fieldName) {
             result[fieldName] = subObj;
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -1,201 +1,157 @@
 'use strict';
 
 
-var HyperSwitch = require('hyperswitch');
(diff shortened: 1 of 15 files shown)


Pchelolo:
Occasionally I've made this PR dependent on #4, so pls review that one first, after that I would rebase this one.

lib/rule.js
```diff
@@ -1,201 +1,157 @@
...
+            return `${obj}.test(${name})`;
         }
-        if (obj[0] !== '/' && obj[obj.length - 1] !== '/') {
-            // not a regex, quote the string
-            result[fieldName] = "'" + obj + "'";
-            return name + " === '" + obj + "'";
+
+        // this is an object, we need to split it into components
+        const subObj = fieldName ? {} : result;
+        const test = Object.keys(obj).map((key) =>
```
d00rman: I think this would be more readable:
```js
Object.keys(obj).map(
  (key) => this._compileMatch(obj[key], subObj, `${name}['${key}']`, key)
).join(' && ');
```

test/feature/rule.js
```diff
@@ -57,12 +57,16 @@ describe('Rule', function() {
         });
 
         it('simple value match', function() {
-            var r = new Rule('rule', {
-                topic: 'nono',
-                exec: {uri: 'a/b/c'},
-                match: {number: 1, string: 'oolala'}
-            });
-            assert.ok(r.test(msg), 'Expected the rule to match the given message!');
+            try {
```
d00rman: Why? If it fails, it should fail loudly

Pchelolo: @d00rman oh, a leftover from development..

Pchelolo:
@d00rman all done

PR #4, opened 2016-04-14
Set up infrastructure for testing with kafka locally and in travis

Pchelolo:
Added infrastructure for running tests with kafka locally and in travis.

This PR conflicts a bit with the others, so this one should go first.

cc @wikimedia/services


README.md
```diff
@@ -2,3 +2,34 @@
...
+mkdir -p $KAFKA_HOME && tar xzf kafka.tgz -C $KAFKA_HOME --strip-components 1
+echo "KAFKA_HOME=$KAFKA_HOME" >> ~/.bash_profile
+echo "PATH=\$PATH:\$KAFKA_HOME/bin" >> ~/.bash_profile
+```
+
+Also, you need to enable topic deletion so that the test scripts could clean up
+kafka state before each test run:
+
+```bash
+echo 'delete.topic.enable=true' >> KAFKA_HOME/config/server.properties
```
d00rman: Why is this needed?

.travis.yml
```diff
@@ -1,11 +1,20 @@
...
 env:
-  - CXX=g++-4.8
+  - KAFKA_HOME=../kafka CXX=g++-4.8
+
+before_install:
+- wget http://www.us.apache.org/dist/kafka/0.8.2.2/kafka_2.10-0.8.2.2.tgz -O kafka.tgz
+- mkdir -p ${KAFKA_HOME} && tar xzf kafka.tgz -C ${KAFKA_HOME} --strip-components 1
+- echo 'delete.topic.enable=true' >> ${KAFKA_HOME}/config/server.properties
+- nohup bash -c "cd ${KAFKA_HOME} && bin/zookeeper-server-start.sh config/zookeeper.properties &"
+- nohup bash -c "cd ${KAFKA_HOME} && bin/kafka-server-start.sh config/server.properties &"
```
d00rman: Put at least `sleep 1` in-between the two, Zookeeper usually needs some time to come up.

README.md
d00rman: Ah, because of the topic dropping. I wonder if this is needed at all

Pchelolo: @d00rman it does't actually drop the topics if you don't have this line.

.travis.yml
```diff
@@ -1,11 +1,21 @@
 language: node_js
+sudo: false
```
gwicke: I *think* this is not needed since this repo / travis test was created after the switch to containers-by-default. But, it certainly doesn't hurt.

.travis.yml
```diff
@@ -1,11 +1,21 @@
...
+  - KAFKA_HOME=../kafka CXX=g++-4.8
+
+before_install:
+- wget http://www.us.apache.org/dist/kafka/0.8.2.2/kafka_2.10-0.8.2.2.tgz -O kafka.tgz
+- mkdir -p ${KAFKA_HOME} && tar xzf kafka.tgz -C ${KAFKA_HOME} --strip-components 1
+- echo 'delete.topic.enable=true' >> ${KAFKA_HOME}/config/server.properties
+- nohup bash -c "cd ${KAFKA_HOME} && bin/zookeeper-server-start.sh config/zookeeper.properties &"
+- sleep 1
+- nohup bash -c "cd ${KAFKA_HOME} && bin/kafka-server-start.sh config/server.properties &"
+- sleep 5
```
gwicke: An alternative to these sleeps could be an `nc` check, similar to what we are doing for Cassandra in the RB tests.

README.md
```diff
@@ -2,3 +2,34 @@
 A [RESTBase](https://github.com/wikimedia/restbase) queuing module for
 [Apache Kafka](http://kafka.apache.org/)
 
+
+## Testing
+
+For testing locally you need to setup and start Apache Kafka and set the 
+`KAFKA_HOME` environment variable to point to the Kafka home directory.
+Here's a sample script you need to run:
```
gwicke: Would it make sense to copy these lines to a script & call it from `npm test`?

Pchelolo: @gwicke not really. I assume that when you're doing some development on change prop locally you already have a running instance of Kafka and Zookeeper, so an attempt to start them up again will fail.

If we restart kafka and zookeeper on each test run - it's just increasing time needed to tun the tests. We don't start cassandra for RB tests after all..

.travis.yml
Pchelolo: @gwicke what `nc` check?

test/feature/static_rules.js
```diff
@@ -0,0 +1,49 @@
...
+            return producer.createTopicsAsync([ 'test_topic_simple_test_rule' ], false)
+        })
+        .then(function() {
+            return changeProp.start();
+        });
+    });
+
+
+    it('Should call simple executor', function() {
+        var service = nock('http://mock.com', {
```
gwicke: I wonder if http://mock.com/ gets a lot of test requests ;)

.travis.yml
gwicke: Sorry, we don't need / have that any more since switching to the `service` stanza. I meant https://github.com/wikimedia/restbase/commit/beb69abf2e6447d159b5fdf243d8f4efa173f9d7

README.md
gwicke: Okay, so we should keep `npm test` untouched. Perhaps we could still move the logic to a shell script, and either tell people to optionally call that, and/or hook it up to some other npm script command (ex: `npm run start-kafka`)?

test/feature/static_rules.js
Pchelolo: hehe, at least our requests never reach it

gwicke:
:+1: overall. And +100 for getting testing started!

Pchelolo:
@gwicke lemme make script for start/stop kafka, one moment

README.md
Pchelolo: @gwicke done

PR #6, opened 2016-04-18
Retry queues and explicit kafka commit

Pchelolo:
1. Retry queues to retry specific rules. For every rule a retry topic is created. In case rule execution results in an error we wrap an original event in an envelope, add a retryCount property and post the event to a special topic. Listener of this topic checks whether retryCount was not exceeded and tries to execute a rule one more time.
2. Explicit kafka commit - only commit if the event handling succeeded or the retry was posted. Switched off `autoCommit` for consumers, so now we commit the offset on after either a rule was successfully executed, or an event was posted in the retry topic.
3. Refactor Role to make most of the private functions actually module-private.

cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -4,17 +4,68 @@
 const HyperSwitch = require('hyperswitch');
 const Template = HyperSwitch.Template;
 
+function _getMatchObjCode(obj) {
+    let code = '{';
+    if (obj.constructor !== Object) {
+        return '';
+    }
+    Object.keys(obj).forEach((key) => {
+        const field = obj[key];
+        let fieldCode = key + ': ';
+        if (field.constructor === Object) {
+            fieldCode += _getMatchObjCode(field);
+        } else {
+            fieldCode += field;
+        }
+        if (code.length > 1) {
+            code += ',';
+        }
+        code += fieldCode;
+    });
+
+    return code + '}';
+}
+
+function _compileMatch(obj, result, name, fieldName) {
+    if (obj.constructor !== Object) {
+        if (typeof obj !== 'string') {
+            // not a string, so it has to match exactly
+            result[fieldName] = obj;
+            return `${name} === ${obj}`;
+        }
+        if (obj[0] !== '/' && obj[obj.length - 1] !== '/') {
+            // not a regex, quote the string
+            result[fieldName] = `'${obj}'`;
+            return `${name} === '${obj}'`;
+        }
+        // it's a regex, we have to the test the arg
+        result[fieldName] = `${obj}.exec(${name})`;
+        return `${obj}.test(${name})`;
+    }
+
+    // this is an object, we need to split it into components
+    const subObj = fieldName ? {} : result;
+    const test = Object.keys(obj).map(
+        (key) => _compileMatch(obj[key], subObj, `${name}['${key}']`, key))
+    .join(' && ');
+    if (fieldName) {
+        result[fieldName] = subObj;
+    }
+    return test;
+}
+
 class Rule {
     constructor(name, spec) {
         this.name = name;
         this.spec = spec || {};
-        this.exec = [];
-        this.topic = '';
-        this.match = {
-            all: true
-        };
-        this.noop = false;
-        this._processRule();
+
+        this.topic = this.spec.topic;
+        if (!this.topic) {
+            throw new Error(`No topic specified for rule ${this.name}`);
+        }
+
+        this.exec = this._processExec(this.spec.exec);
+        this._match = this._processMatch(this.spec.match);
     }
 
     /**
@@ -24,7 +75,7 @@ class Rule {
      * @return true if no match is set for this rule or if the message matches
      */
     test(message) {
-        return this.match.all || this._matchTest(message);
+        return !this._match || !this._match.test || this._match.test(message);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


d00rman:
I have a concern regarding the creation and subsequent subscription to retry queues. The general approach we are taking is that topics are static, which means that retry queues can easily be inferred from them, hence no need for their dynamic creation. Another question is: should the process subscribed to the original topic handle retry ones as well? IMHO, it'd be more efficient and resilient to have a different process handle that (note that, in our usage, the number of change-prop processes greatly outnumbers the number of topics and rules). Going even further, because all of the retry topics have effectively the same schema, should we just have a single retry topic? Honestly, I'm not convinced by the latter, but conceptually it does sound tempting.

Pchelolo:
> Should the process subscribed to the original topic handle retry ones as well?

I'm not sure I understand how dynamic topic creation imply that the same worker is used for a main topic and for a retry topic? The consumer group of a retry topic is different from a consumer group of a main topic, so normal balancing will be applied, the same way as for other rules/topics. The call to `createTopics` just ensures the topics are there, in 99% of cases it's a no-op. Basically proposed scheme is identical to static topics, bc after the service was first started all the retry topics are there.

> Retry queues can easily be inferred from them, hence no need for their dynamic creation.

This means we will have one worker handling retries from all the rules for the topic. So it would be actually worse for load distribution then the proposed scheme.

> Going even further, because all of the retry topics have effectively the same schema, should we just have a single retry topic? Honestly, I'm not convinced by the latter, but conceptually it does sound tempting.

This would imply that a single worker will have to run retries from all of the other rules/topics, which is bad. Or we will need to have multiple workers subscribe to it with different consumer groups and filter-out retries - not good either.

lib/rule_executor.js
```diff
@@ -22,58 +22,131 @@ class RuleExecutor {
...
+            this.retryConsumer = this._setConsumerLoggers(consumer, this.rule.name, retryTopicName);
+            this.retryConsumer.on('message', (msg) => {
+                return P.try(() => {
+                    const message = this._safeParse(msg.value);
+                    if (!message) {
+                        // Don't retry if we can't parse an event, just log.
+                        return;
+                    }
+
+                    if (message.retryCount >= (this.rule.spec.retry_limit || 3)) {
```
d00rman: For readability, it'd be better to have `3` as a module-level constant.

lib/rule_executor.js
```diff
@@ -22,58 +22,131 @@ class RuleExecutor {
...
+                        });
+                    }
+                })
+                .then(() => this.retryConsumer.commitAsync());
+            });
+        });
+    }
+
+    _retry(retryMessage) {
+        return P.delay(this.rule.spec.retry_delay || 500)
```
d00rman: Idem (re constant)

lib/rule_executor.js
```diff
@@ -22,58 +22,131 @@ class RuleExecutor {
...
+        .then(() => this.retryProducer.sendAsync([{
+            topic: this._getRetryTopicName(),
+            messages: [ JSON.stringify(retryMessage) ]
```
d00rman: This suggests we are leaving room for batching retries, but the above message-handling code doesn't actually take that into account. `this._exec` uses the message's value directly, but here we set it as an array.

d00rman: Grr, it's the producer interface. Pardon my ignorance.

d00rman:
>> Should the process subscribed to the original topic handle retry ones as well?

> I'm not sure I understand how dynamic topic creation imply that the same worker is used for a main topic and for a retry topic? The consumer group of a retry topic is different from a consumer group of a main topic, so normal balancing will be applied, the same way as for other rules/topics. The call to createTopics just ensures the topics are there, in 99% of cases it's a no-op. Basically proposed scheme is identical to static topics, bc after the service was first started all the retry topics are there.

I gave it a second look and actually it's ok. I was too tired yesterday when writing the comment.

>> Retry queues can easily be inferred from them, hence no need for their dynamic creation.

> This means we will have one worker handling retries from all the rules for the topic. So it would be actually worse for load distribution then the proposed scheme.

The idea would actually be to have one retry topic per regular topic with an extra `rule` parameter which would indicate the rule to execute. Then, clients would connect using the `${topic}-${rule}` consumer group ID, which would ensure distinct processes handle retry messages for different rules.

Wrt to static vs dynamic retry topic creation, I do agree that these are mostly change-propagation-specific. However, with the above suggestion these could generalised per-topic and other consumers could also make use of them. A bigger concern here, though, is the fact that having this dynamic creation means that we'd need to manage the retry topics ourselves when faced with dynamic rule (un)subscription, which I don't think it's worth the trouble.

Pchelolo:
@d00rman 

> The idea would actually be to have one retry topic per regular topic with an extra rule parameter which would indicate the rule to execute. Then, clients would connect using the ${topic}-${rule} consumer group ID, which would ensure distinct processes handle retry messages for different rules.

Then all of the retry consumers for the given topic will receive all of the messages for all rules, so they will need to filter out most of the rules. This adds unnecessary load both on the broker and on the change-prop worker. With a per-rule retry topic consumer only gets messages that it needs to process.

>  However, with the above suggestion these could generalised per-topic and other consumers could also make use of them.

Consumers shouldn't really bother with the retry topics, as they just have to return an HTTP error and change propagation will figure out the retry strategy - it's change-prop's job to do that. 

> A bigger concern here, though, is the fact that having this dynamic creation means that we'd need to manage the retry topics ourselves when faced with dynamic rule (un)subscription, which I don't think it's worth the trouble.

Great point with UNsubscribtion, as normally we'd need to delete a retry topic for that rule. But as deleting topics is not supported a garbage topic will be left there forever. This is indeed a problem, but I'm not 100% sure whether it's serious enough.

d00rman:
On Thursday, 21 April 2016, Petr Pchelko <notifications@github.com> wrote:
> @d00rman
>
> The idea would actually be to have one retry topic per regular topic with
an extra rule parameter which would indicate the rule to execute. Then,
clients would connect using the ${topic}-${rule} consumer group ID, which
would ensure distinct processes handle retry messages for different rules.
>
> Then all of the retry consumers for the given topic will receive all of
the messages for all rules, so they will need to filter out most of the
rules. This adds unnecessary load both on the broker and on the change-prop
worker. With a per-rule retry topic consumer only gets messages that it
needs to process.

I don't honestly expect filtering out to be a real problem in practice;
deducing a message needs to be discarded is very cheap when compared to
actually executing it. Also, I wouldn't worry too much seeing that most of
our workers will be idling :)

I don't get your point about the brokers, though.

>
> However, with the above suggestion these could generalised per-topic and
other consumers could also make use of them.
>
> Consumers shouldn't really bother with the retry topics, as they just
have to return an HTTP error and change propagation will figure out the
retry strategy - it's change-prop's job to do that.

Not consumers of the change-prop service, but other kafka consumers.

>
> A bigger concern here, though, is the fact that having this dynamic
creation means that we'd need to manage the retry topics ourselves when
faced with dynamic rule (un)subscription, which I don't think it's worth
the trouble.
>
> Great point with UNsubscribtion, as normally we'd need to delete a retry
topic for that rule. But as deleting topics is not supported a garbage
topic will be left there forever. This is indeed a problem, but I'm not
100% sure whether it's serious enough.

The EventBus system is not our private kafka install, but an org-wide
system and we should be mindful about what we do there. So, yes, it's an
issue.

>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub<
https://ci3.googleusercontent.com/proxy/EXZSqPKL9IZAj1YfZ3FpuCDiOj9-10821CYf-oiJ8wpmvhtDn0vM0Fw4sNFm879wNYFgv9btX6Dr-ZsStF3Gm39c5nnM5olGlhqMyVC-DnjjMfkqIKeFIac4PSMfBvl61PHgG8NAuQAaRlV5ffGEmy0oET518g=s0-d-e1-ft#https://github.com/notifications/beacon/AJvLfZjA5bLgFpjSFiiGERBl8TIgnFCuks5p57Q5gaJpZM4IJ6zj.gif
>

-- 
Marko Obrovac, PhD
Senior Services Engineer
Wikimedia Foundation

Pchelolo:
@d00rman Ok, I'm convinced :) I'll rework this patch to use static per-topic retry queues. 

So we need to create a schema for a retry topic and make the event logging-service create retry topics transparently.

Other question is whether we want to push messages to a retry topic directly from change-prop, bypassing the event logging service, or if we want to go through it. What do you think?

d00rman:
> Other question is whether we want to push messages to a retry topic directly from change-prop, bypassing the event logging service, or if we want to go through it. What do you think?

Given the circumstances I think it's OK to send them directly to Kafka.

lib/rule_executor.js
```diff
@@ -22,58 +26,180 @@ class RuleExecutor {
...
+            meta: {
+                topic: this._retryTopicName(),
+                schema_uri: 'retry/1',
+                uri: event.meta.uri,
+                request_id: event.meta.request_id,
+                id: uuid.fromDate(now),
+                dt: now.toISOString(),
+                domain: event.meta.domain
+            },
+            consumer_id: this._consumerId(),
```
d00rman: `s/consumer_id/emitter_id`

lib/rule_executor.js
```diff
@@ -22,58 +26,180 @@ class RuleExecutor {
...
+                .then(() => this.retryConsumer.commitAsync());
             });
-            consumer.on('error', (err) => this.log(`warn/error/${rule.name}`, {
-                err,
-                rule: logRule
-            }));
         });
     }
 
+    _shouldRetry(errorRes) {
```
d00rman: It would be nice if we could have the same functionality as in HyperSwitch for status codes, so that we can specify:

```yaml
retry_on: ['50x']
```

and friends.

d00rman: Also, it'd be good to have a check for `retry_left <= 0` (possibly in here given the method's name). As it currently stands, for each failed message, we are always enqueuing and dequeuing one extra message that will never be executed (the one which has `retry_left: 0`).

d00rman:
Some minor bits and pieces and I think we'll be good to go!

lib/rule.js
```diff
@@ -1,20 +1,118 @@
...
 
-
+const stringify = require('json-stable-stringify');
 const HyperSwitch = require('hyperswitch');
 const Template = HyperSwitch.Template;
 
+/**
+ * Creates a JS function that verifies property equality
+ *
+ * @param catchDefinition the condition in the format of 'catch' and 'return_if' stanza
```
d00rman: Copy/paste artefact?

lib/rule.js
```diff
@@ -1,20 +1,118 @@
...
+ * @returns {Function} a function that verifies the condition
+ */
+function _compileCatchFunction(catchDefinition) {
```
d00rman: This was obviously c/p from HyperSwitch, so I'd really prefer it if the variable names and error messages were adjusted.

lib/rule.js
```diff
@@ -1,20 +1,118 @@
...
+
+        this.topic = this.spec.topic;
+        if (!this.topic) {
+            throw new Error(`No topic specified for rule ${this.name}`);
+        }
+
+        if (this.spec.retry_on) {
+            this.shouldRetry = _compileCatchFunction(this.spec.retry_on);
+        } else {
+            this.shouldRetry = function() {
```
d00rman: Would it be beneficial to have the default retry condition be set on `['50x']` ? I can't think of situation where that would not be the case.

Pchelolo:
@d00rman Updated according to your latest comments

gwicke:
FWIW, I'm a bit sceptical about having a single retry topic per primary topic, rather than per rule. For huge topics like resource_changed with lots of traffic and subscribers, this could theoretically lead to significant wasted effort. In the worst case, this could lead to a chain reaction, with trouble elsewhere triggering high change propagation load (from all subscribers going through all retries per topic).

But, if this theoretical possibility turns out to be a problem in practice, we can always switch back to more targeted per-rule retry topics. As all of this retry machinery is private & namespaced to the retry service, none of the consumers would notice.

PR #7, opened 2016-04-20
Support producing messages to topics from a rule executor

Pchelolo:
See https://phabricator.wikimedia.org/T133221 for some background. 

I'm not 100% sure we really want to allow this or not (should all the event production go through the eventlogging-service, or would we allow doing shortcuts like this?)

cc @wikimedia/services

### DIFF
--- a/sys/kafka.js
+++ b/sys/kafka.js
@@ -7,6 +7,9 @@
 
 
 const P = require('bluebird');
+const HTTPError = require('hyperswitch').HTTPError;
+const uuid = require('cassandra-uuid').TimeUuid;
+
 const Rule = require('../lib/rule');
 const KafkaFactory = require('../lib/kafka_factory');
 const RuleExecutor = require('../lib/rule_executor');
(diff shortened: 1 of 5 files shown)


Pchelolo:
Tests depend on https://github.com/wikimedia/swagger-router/pull/40

gwicke:
I'm :+1: on the general approach.

We discussed making the interface more similar to the eventbus proxy service:

- use an `events` entry point
- per-event meta element, specifying the topic & other bits

One potential issue we found is that the eventbus proxy requires meta elements like `dt` to be supplied, but doing so in each rule handler would get quite repetitive. We might want to fill in several of those meta properties in this end point, so that rule handlers only need to specify the essentials.

Pchelolo:
@gwicke Ye, I've updated it

sys/kafka.js
```diff
@@ -48,14 +93,21 @@ module.exports = (options) => {
...
             }
         },
         operations: {
-            setup_kafka: kafkaMod.setup.bind(kafkaMod)
+            setup_kafka: kafkaMod.setup.bind(kafkaMod),
+            produce: kafkaMod.produce.bind(kafkaMod)
         },
         resources: [{
-            uri: '/{domain}/sys/queue/setup'
+            uri: '/sys/queue/setup'
```
d00rman: Why is the domain being removed here?

test/feature/static_rules.js
```diff
@@ -18,13 +20,18 @@ describe('Basic rule management', () => {
     let producer;
     let retrySchema;
 
-    before(() => {
+    before(function() {
```
d00rman: `before(() => { ...` ?

test/mocha.opts
```diff
@@ -1,2 +1 @@
 --recursive
---no-timeouts
```
d00rman: Why are we removing this?

test/feature/static_rules.js
Pchelolo: No. `mocha` assigned some properties to the `function` we provide it, so that we can call `this.timeout(0)`. Arrow functions don't have there own `this`, so the timeout thingy doesn't work.

sys/kafka.js
Pchelolo: @d00rman Why do we need it at all?

test/mocha.opts
Pchelolo: Because I have set up the timeouts and we want the tests to timeout. When the development is happening some tests could get into a bad state, and some of them can wait indefinitely without a timeout.

sys/kafka.js
d00rman: I completely missed that it was removed from the top declaration as well :)

PR #8, opened 2016-04-26
Set up config parameter to parameterize production and consumption DC

Pchelolo:
Two changes:

1. We require the user to set up config parameters for produce_dc and consume_dc - it's used as a topic prefix.
2. All event production now go through the production endpoint.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -12,8 +12,10 @@ class KafkaFactory {
      * @param {Object} kafkaConf Kafka connection configuration
      * @param {string} kafkaConf.uri Zookeper URI with host and port
      * @param {string} kafkaConf.clientId Client identification string
-     * @param {string} kafkaConf.consume_dc DC name to consume from
-     * @param {string} kafkaConf.produce_dc DC name to produce to
+     * @param {string} [kafkaConf.consume_dc] DC name to consume from
+     * @param {string} [kafkaConf.produce_dc] DC name to produce to
+     * @param {string} [kafkaConf.dc_name] DC name to use both for
+     *                                     production and consumption
      * @constructor
      */
     constructor(kafkaConf) {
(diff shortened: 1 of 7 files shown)


lib/kafka_factory.js
```diff
@@ -12,10 +12,24 @@ class KafkaFactory {
...
      * @constructor
      */
     constructor(kafkaConf) {
         this.kafkaConf = kafkaConf;
+
+        if (!this.kafkaConf.uri) {
+            throw new Error('uri config parameter is required by kafka config');
+        }
+
+        if (!this.kafkaConf.consume_dc) {
```
d00rman: This is rather WMF-specific, so I think not having these options should be tolerable (as in, if they're not given, just don't prefix the topic names). Also, how about adding a third option, `dc_name` that would be used for both if specified?

d00rman:
Minor comment, otherwise :+1: from me

Pchelolo:
@d00rman Updated.

lib/kafka_factory.js
gwicke: Multi-DC handling isn't really WMF-specific, but you are right that we should make it easy to start out with a single DC. We decided to follow Cassandra's lead of providing a default DC name: `datacenter1`. This is already implemented in the eventbus proxy, so lets do the same here.

sys/kafka.js
```diff
@@ -19,7 +19,9 @@ class Kafka {
         this.log = options.log || function() { };
         this.kafkaFactory = new KafkaFactory({
             uri: options.uri || 'localhost:2181/',
-            clientId: options.client_id || 'change-propagation'
+            clientId: options.client_id || 'change-propagation',
+            consume_dc: options.consume_dc,
+            produce_dc: options.produce_dc
```
d00rman: `dc_name` seems to be missing in this list.

lib/kafka_factory.js
```diff
@@ -69,10 +63,11 @@ class KafkaFactory {
      * @returns {Promise} a promise that's resolved when a consumer is ready
      */
     newConsumer(client, topic, groupId, offset) {
+        const prefixedTopic = this.consumeDC ? `${this.consumeDC}.${topic}` : topic;
```
gwicke: We should use `datacenter1` as the default DC.

PR #9, opened 2016-04-26
Batched processing of the backlinks

Pchelolo:
Missing bits and pieces:
1. Need to set up a continue schema in EventBus

cc @wikimedia/services

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -8,6 +8,12 @@ services:
     conf:
       port: 7273
       spec:
+        x-sub-request-filters:
+          - type: default
+            name: http
+            options:
+              allow:
+                - pattern: /^https?:\/\//
         title: The Change Propagation root
         paths:
           /sys/links:
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -45,21 +45,24 @@ services:
                           derived_field: '{{message.message}}'
 
                     process_redlinks:
-                      topic: revision_create
+                      topic: mediawiki.revision_create
                       retry_limit: 0
+                      match:
+                        rev_parent_id: undefined
                       exec:
                         method: post
-                        uri: '/sys/links/backlinks/{{message.title}}'
-                        body: '{{$.message}}'
+                        uri: '/sys/links/backlinks/{message.title}'
+                        body: '{{globals.message}}'
 
                     rerender_restbase:
                       topic: resource_change
                       match:
                         meta:
                           uri: '/https:\/\/([a-zA-Z0-9\.]+)\/wiki\/(.+)/'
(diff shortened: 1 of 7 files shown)


config.test.yaml
```diff
@@ -38,6 +44,25 @@ services:
                           test_field_name: test_field_value
                           derived_field: '{{message.message}}'
 
+                    process_redlinks:
+                      topic: revision_create
+                      retry_limit: 0
+                      exec:
+                        method: post
+                        uri: '/sys/links/backlinks/{{message.title}}'
```
gwicke: You probably want to use single quotes here, to make sure that the title is properly escaped.

config.test.yaml
```diff
@@ -38,6 +44,25 @@ services:
...
+                        body: '{{$.message}}'
+
+                    rerender_restbase:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/https:\/\/([a-zA-Z0-9\.]+)\/wiki\/(.+)/'
+                      exec:
+                        method: get
+                        uri: 'https://{{$.match.meta.uri[1]}}/api/rest_v1/page/html/{{$.match.meta.uri[2]}}'
```
gwicke: Similar here, especially for the second match part.

config.test.yaml
```diff
@@ -38,6 +44,25 @@ services:
...
+                      exec:
+                        method: post
+                        uri: '/sys/links/backlinks/{{message.title}}'
+                        body: '{{$.message}}'
+
+                    rerender_restbase:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/https:\/\/([a-zA-Z0-9\.]+)\/wiki\/(.+)/'
```
gwicke: To avoid duplicate processing in the transition period, it would be nice to add a tag match, limited to changeprop-produced events.

sys/backlinks.js
```diff
@@ -0,0 +1,120 @@
...
+    setup(hyper) {
+        return hyper.post({
+            uri: '/sys/queue/subscriptions',
+            body: {
+                backlinks_continue: {
+                    topic: CONTINUE_TOPIC_NAME,
+                    exec: [
+                        {
+                            method: 'post',
+                            uri: '/sys/links/backlinks/{{message.original_event.title}}',
```
gwicke: Single braces would be better for escaping.

gwicke:
LGTM overall. A couple of nits inline.

config.test.yaml
Pchelolo: @gwicke done

gwicke:
:+1:

PR #10, opened 2016-04-27
Rules: Add the `match_not` predicate (release v0.2.0)

d00rman:
There are situations where we might need to negate the outcome of a rule match test. One such example is updating the summary endpoint for non-wiktionaries. This PR adds the functionality via the `match_not` predicate. If it is present, its condition is evaluated and negated and the message is accepted only if it doesn't match it. This works rather well in conjuction with the `match` predicate: if both are specified, they both need to be evaluated positively in order for the message to be accepted. Note that `match_not` is used only for testing messages, it cannot be used for rule execution expansion.

cc @wikimedia/services

### DIFF
--- a/test/feature/rule.js
+++ b/test/feature/rule.js
@@ -113,6 +113,45 @@ describe('Rule', function() {
             }, Error);
         });
 
+        it('match_not', function() {
+            var r = new Rule('rule', {
+                topic: 'nono',
+                exec: {uri: 'a/b/c'},
+                match_not: {meta: {uri: '/my-url/'}}
+            });
+            assert.ok(r.test(msg), 'Expected the rule to match the given message!');
+        });
+
+        it('matches match and match_not', function() {
+            var r = new Rule('rule', {
+                topic: 'nono',
+                exec: {uri: 'a/b/c'},
+                match: {number: 1},
+                match_not: {meta: {uri: '/my-url/'}}
+            });
+            assert.ok(r.test(msg), 'Expected the rule to match the given message!');
+        });
+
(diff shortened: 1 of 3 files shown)


test/feature/rule.js
```diff
@@ -113,6 +113,45 @@ describe('Rule', function() {
...
+            });
+            assert.ok(r.test(msg), 'Expected the rule to match the given message!');
+        });
+
+        it('matches match but not match_not', function() {
+            var r = new Rule('rule', {
+                topic: 'nono',
+                exec: {uri: 'a/b/c'},
+                match: {number: 1},
+                match_not: {meta: {uri: '/fake/'}}
```
Pchelolo: This predicate will actually match, so the test is failing

Pchelolo:
May be better `not_match`?

gwicke:
I'm wondering how far we should push this custom match DSL, rather than supporting functions / expressions as in regular request handlers.

d00rman:
> May be better `not_match`?

I'm not really happy with `match_not` either, I was in a dilemma between `match_not` and `dont_match` but went for the former because it has a note of poetry in it while still being an imperative :)

> I'm wondering how far we should push this custom match DSL, rather than supporting functions / expressions as in regular request handlers.

I don't find `match_not` too sexy either, but I think it's more convenient than having functions / expressions for that (IMHO, it's more readable like this). But in general I agree that if start adding stuff like crazy to it. Maybe revisit this in the future if/when we find features that are needed but lacking?

Pchelolo:
👍

PR #11, opened 2016-04-27
Produce events to error topic

Pchelolo:
If retries failed or message parsing failed - push a fatal error to the error topic.

The topic schema PR: https://gerrit.wikimedia.org/r/#/c/284986/

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -222,7 +222,7 @@ class RuleExecutor {
      * @param {string|Object} event an original event. In case JSON parsing failed - it's a string.
      */
     _constructErrorMessage(e, event) {
-        const eventUri = typeof event === 'string' ? 'unknown' : event.meta.uri;
+        const eventUri = typeof event === 'string' ? '/error/uri' : event.meta.uri;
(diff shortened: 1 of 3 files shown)


lib/rule_executor.js
```diff
@@ -67,9 +68,13 @@ class RuleExecutor {
 
     _safeParse(message) {
         try {
-            return JSON.parse(message);
+            return P.resolve(JSON.parse(message));
         } catch (e) {
-            this.log(`error/${this.rule.name}`, e);
+            return this.hyper.post({
```
gwicke: Should we log these failures as well?

gwicke:
LGTM, only one minor question inline.

d00rman:
LGTM too, but we cannot merge this before https://gerrit.wikimedia.org/r/#/c/284986/ is resolved and merged.

Pchelolo:
Also depends on #32 since we need to add `triggered_by` field to the error event

ottomata:
Merged https://gerrit.wikimedia.org/r/#/c/284986/

PR #12, opened 2016-04-27
Create sample config for summary and definition updates

Pchelolo:
Basically it's a loop since the config will live in puppet, but let's have our local copy too for review and testing.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -1,0 +1,73 @@
+num_workers: 0
+logging:
+  name: changeprop
+  level: info
+services:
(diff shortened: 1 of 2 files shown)


config.example.yaml
```diff
@@ -1,48 +0,0 @@
-# Number of worker processes to spawn.
```
d00rman: Why are we removing the example config? It should serve for people outside of WM.

Pchelolo: Oh, ye, I wanted to duplicate the file but renamed instead..

PR #14, opened 2016-04-29
Backlinks: Allow users to specify a full MW API request template

d00rman:
In production, we call the MW API directly, bypassing nginx and Varnish. In order for that to work, we must set the `host` header in the request. This PR allows the user to set the full MW API request template, and then it extends it with the method and body used to fulfil the backlinks request.

Also: set the port in the config to 7272 as that's what we are using in production.

cc @wikimedia/services

### DIFF
--- a/sys/backlinks.js
+++ b/sys/backlinks.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const extend = require('extend');
 const HyperSwitch = require('hyperswitch');
 const Template = HyperSwitch.Template;
 
@@ -8,9 +9,8 @@ const CONTINUE_TOPIC_NAME = 'change-prop.backlinks.continue';
 class BackLinksProcessor {
     constructor(options) {
         this.options = options;
-        this.backLinksRequest = new Template({
+        this.backLinksRequest = new Template(extend(true, {}, options.templates.mw_api, {
             method: 'post',
-            uri: options.templates.apiURITemplate,
             body: {
                 format: 'json',
                 action: 'query',
@@ -20,7 +20,7 @@ class BackLinksProcessor {
                 blcontinue: '{{message.continue}}',
                 bllimit: 500
             }
-        });
+        }));
     }
 
     setup(hyper) {
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -21,7 +21,10 @@ services:
               - path: sys/backlinks.js
                 options:
                   templates:
-                    apiURITemplate: 'https://{{message.meta.domain}}/w/api.php'
+                    mw_api:
+                        uri: 'https://{{message.meta.domain}}/w/api.php'
+                        headers:
+                          host: '{{message.meta.domain}}'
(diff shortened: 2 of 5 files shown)


sys/backlinks.js
```diff
@@ -8,9 +9,8 @@ const CONTINUE_TOPIC_NAME = 'change-prop.backlinks.continue';
 class BackLinksProcessor {
     constructor(options) {
         this.options = options;
-        this.backLinksRequest = new Template({
+        this.backLinksRequest = new Template(extend(true, {}, options.templates.mw_api, {
```
gwicke: [minor] Would the standard Object.assign be sufficient here?

gwicke:
:+1: from me overall, only one minor question inline.

sys/backlinks.js
Pchelolo: `new Template` already clone the spec prior to using it, so this is not needed at all. Neither `extend` nor `Object.assign`. Although template makes a shallow copy, but if we wanna fix it, we need to fix it there, not here.

Pchelolo: Amend number 2. I understand what @d00rman is doing here and I support him. A deep extend will merge object properties instead of overriding them. Eg:
```
extend(true, {a: {a: 'a'}}, {a: {b: 'b'}}) // => { a: { a: 'a', b: 'b' } }
Object.assign({a: {a: 'a'}}, {a: {b: 'b'}}) // => { a: { b: 'b' } }
```
So use of extend allows to partially specify the body and/or set some headers here in the template. So I support @d00rman on this one.

Pchelolo:
LGTM, although it would be nice to cover it with tests by adding expected `reqheaders` [here](https://github.com/wikimedia/change-propagation/blob/master/test/feature/static_rules.js#L84)

Pchelolo:
Also, I think the same will be true for `restbase`, so we might wanna add the host header [here](https://github.com/wikimedia/change-propagation/blob/master/config.example.wikimedia.yaml#L53) and [here](https://github.com/wikimedia/change-propagation/blob/master/config.example.wikimedia.yaml#L71) right away, and also the tests for them, so that the prod config is as similar to the example one as possible

sys/backlinks.js
gwicke: Currently, the only nested object is the body, which looks less likely as a customization target. In any case, I don't have strong feelings about this, just wondered why we need a deep clone here.

d00rman: I agree that given the special case of the template, it is highly unlikely that one would specify the body as well, but I opted for extending it simply because of correctness. There is no effective overhead to doing so, since the template is created only once per process on start-up.

d00rman:
@Pchelolo according to nock's documentation, *If reqheaders is not specified or if host is not part of it, Nock will automatically add host value to request header.*

Pchelolo:
@d00rman right, right... I've just checked that it the test will not fail regardless of whether we have this patch or not. 

I'm +1 on merging this one

PR #15, opened 2016-04-29
Catch TopicNotExistError and report it properly

Pchelolo:
Based on morning discussion with @d00rman - do not crash if topic doesn't exist. 

However, I still think we need to crash the service. Non existent topic in a static config means that the service is misconfigured, so we should take it down. As workers do not load updated config when they're respawned any more, on config deployment we will immediately notice that something is wrong, because rolling restart will fail on the first node, while if we swallow an error and just log it, all nodes will be restarted before we notice an error in logstash, thus putting the whole change prop cluster in a loose state. So I propose to add `process.exit(1)` in that catch block. 

When we have dynamic subscriptions, this will change, but we can separate the `subscribe` method to `subscribeStatic` and `subscribeDynamic` - the first one will fail on non-existent topic, the second one will just throw, log error and return 400 to the subscription request origin.

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -208,7 +208,10 @@ class RuleExecutor {
                 });
             });
         })
-        .catch(TopicsNotExistError, (e) => this.log('error/topic', e));
+        .catch(TopicsNotExistError, (e) => {
+            this.log('error/topic', e);
+            process.exit(1);
+        });
     }
 }
 
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -210,7 +210,8 @@ class RuleExecutor {
         })
         .catch(TopicsNotExistError, (e) => {
             this.log('error/topic', e);
-            process.exit(1);
+            // Exit async to let the logs get processed.
+            setTimeout(() => process.exit(1), 100);
         });
     }
 }
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -2,6 +2,7 @@
 
 const P = require('bluebird');
 const uuid = require('cassandra-uuid').TimeUuid;
+const TopicsNotExistError = require('wmf-kafka-node/lib/errors').TopicsNotExistError;
 
 const DEFAULT_RETRY_DELAY = 500;
 const DEFAULT_RETRY_LIMIT = 3;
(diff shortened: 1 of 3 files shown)


lib/rule_executor.js
```diff
@@ -206,6 +207,10 @@ class RuleExecutor {
                     .finally(() => this._reportMetrics(statName, startTime, message));
                 });
             });
+        })
+        .catch(TopicsNotExistError, (e) => {
+            this.log('error/topic', e);
```
gwicke: Since logging is async, we should probably delay a little before exiting. This reminds me, we should probably also return a Promise from the logger in service-runner, so that clients can (optionally) wait for logging to finish before exiting.

gwicke:
I agree that failing early & loudly is better than failing later, at runtime. In production, this will immediately abort a deploy, rather than possibly proceeding to take out the entire cluster.

For `subscribeDynamic`, are there use cases where it would not make sense to dynamically create those topics if they are missing?

lib/rule_executor.js
Pchelolo: @gwicke done

Pchelolo:
@gwicke the `subscribeDynamic` is not really in scope of this PR, we will have to make a decision on it later and it's not an easy one. Both dynamic and static topics have their pros and cons, so let have this discussion separately

d00rman:
My main motivation for not failing were exactly dynamic rules. For static rules, I can concede that it makes sense to fail, but failing on dynamic rules would mean that workers would be slowly taken out one by one which is far from desirable.

As for failing, I think this would a case where it would be more desirable to fail the master rather than only the worker. I think that this needs to be fixed in general in service-runner and/or hyperswitch - if there is a failure during start-up don't restart the failing workers over and over, but fail completely. In the current state of affairs, systemd doesn't categorise that as a start-up failure because the master is still up and running.

gwicke:
I would expect the port checks to detect the worker failure.

d00rman:
And they do on deploy, but when you're on the node, `systemctl status X` and `ps -ef | grep X` would both tell you the service is running, which is misleading. We shouldn't make it hard to detect whether the service is up or not, IMHO.

Pchelolo:
@d00rman I can file a ticket for service runner to support a 'dead letter' in master and fail it. After that we can update this.

d00rman:
Cool, thnx! Maybe I'm wrong, but given that we are talking about stateless services, I can't find a case where one worker would be able to start up, while another wouldn't. So I think it makes sense to simply fail the master if the worker fails on start-up.

Pchelolo:
@d00rman filed https://phabricator.wikimedia.org/T134002

PR #16, opened 2016-04-29
Added infrastructure for varnish purging and a test

Pchelolo:
Moving Varnish purging to change-propagation.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -48,7 +48,7 @@ services:
                           - '5xx'
                       match:
                         meta:
-                          uri: '/^(https?):\/\/[a-zA-Z0-9:\.]+\/api\/rest_v1\/page\/html\/([^/]+)$/'
+                          uri: '/^(https?):\/\/[^\/]+\/api\/rest_v1\/page\/html\/([^/]+)$/'
                         tags:
                           - restbase
                       match_not:
@@ -72,6 +72,7 @@ services:
                         meta:
                           # These URIs are coming from RESTBase, so we know that article titles will be normalized
                           # and main namespace articles will not have : (uri-encoded, so %3a or %3A)
+                          uri: '/^(https?):\/\/[^\/]+\/api\/rest_v1\/page\/html\/((?:(?!%3a|%3A).)+)$/'
                           domain: '/wiktionary.org$/'
                         tags:
                           - restbase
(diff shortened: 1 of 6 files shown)


gwicke:
:+1: on the code change from me.

How (in which order) should we deploy this?

Pchelolo:
@gwicke I think that these PRs that add use-cases for change-prop should go one-by-one.

gwicke:
@Pchelolo, another concern for migrating purging is executing those purges twice. Can we target only a subset of purges at first?

Pchelolo:
@gwicke For that we will add a more targeted match in puppet config, but here we need the whole thing right away

d00rman:
Aren't purges supposed to be handled by a separate service (as we discussed with the Performance team) that listens to the `resource_changed` topic?

gwicke:
@d00rman, my understanding is that purge handling in the change propagation service would be a temporary solution, until the separate purge service arrives & takes over. I would not be surprised if that will take some time, and will require some more thinking around XKey & timing of purges.

d00rman:
This doesn't look right to me. The service purges RESTBase content as soon as MW emits a purge event? There is a strong possibility there will be a window of time where the content has been purged, but not yet re-rendered. The correct way would be for it to match `api/rest_v1` routes in the `resource_change` (which are anyway emitted by RESTBase) and then purge that content.

Pchelolo:
@d00rman but that's exactly what's happening (https://github.com/wikimedia/change-propagation/pull/16/files#diff-541b6e195e9da580fc0bd42761ee761fR85)

lib/rule_executor.js
```diff
@@ -199,6 +199,7 @@ class RuleExecutor {
                         return this._exec(message)
                         .catch((e) => {
                             if (this.rule.shouldRetry(e)) {
+                                this.log('warn/retry', e);
```
d00rman: Let's not do that. When part of the infra is down or a similar issue occurs, we'll have a gazillion log entries.

config.example.wikimedia.yaml
```diff
@@ -68,11 +74,25 @@ services:
...
                         # Don't encode title since it should be already encoded
                         uri: '{{match.meta.uri[1]}}://{{message.meta.domain}}/api/rest_v1/page/definition/{{match.meta.uri[2]}}'
                         headers:
                           cache-control: no-cache
+
+                    purge_varnish:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/^https?:\/\/[a-zA-Z0-9:\.]+\/api\/rest_v1\/(.+)$/'
```
d00rman: Domain regexes are really tricky. Let's use `[^\/]+` instead

Pchelolo:
@d00rman I've addressed your comments

PR #17, opened 2016-04-29
Don't rerender definitions for non-main namespaces

Pchelolo:
In RESTBase we don't rerender definitions for non-main namespaces, so we need to do the same in change prop. URIs in events in change-prop come from RB, so we're sure that pages from main namespace will not use any namespace aliases etc, so we can just update a regex to not match `%3A` and `%3a` char sequences.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -64,7 +64,9 @@ services:
                           - '5xx'
                       match:
                         meta:
-                          uri: '/^(https?):\/\/[a-zA-Z0-9\:\.]+\/api\/rest_v1\/page\/html\/([^/]+)/'
+                          # These URIs are coming from RESTBase, so we know that article titles will be normalized
+                          # and main namespace articles will not have : (uri-encoded, so %3a or %3A)
+                          uri: '/^(https?):\/\/[a-zA-Z0-9\:\.]+\/api\/rest_v1\/page\/html\/((?:(?!%3a|%3A).)+)$/'
                           domain: '/wiktionary.org$/'
                       exec:
                         method: get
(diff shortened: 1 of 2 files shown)


config.example.wikimedia.yaml
```diff
@@ -64,7 +64,9 @@ services:
                           - '5xx'
                       match:
                         meta:
-                          uri: '/^(https?):\/\/[a-zA-Z0-9\:\.]+\/api\/rest_v1\/page\/html\/([^/]+)/'
+                          # These URIs are coming from RESTBase, so we know that article titles will be normalized
+                          # and main namespace articles will not have : (uri-encoded, so %3a or %3A)
+                          uri: '/^(https?):\/\/[a-zA-Z0-9\:\.]+\/api\/rest_v1\/page\/html\/((?:(?!%3a|%3A).)+)$/'
```
gwicke: Is the second half of this assertion (no `:` in wiktionary titles) actually correct? I would agree that the number is likely small, but is it really zero?

Pchelolo: Hm.... On one hand, it's possible to have a title with `:` where prefix is not a namespace, however this is limited to wiktionary where every title is a word, so I'm 99.9% sure there is no titles with `:` there. 

Pulling the `mediawiki-title` lib and running a full title parsing is really an overkill for a couple of potentially bad titles. 

As for cache invalidation - we indeed do not invalidate caches for definition endpoint on non-main namespaces. However in practice I think it's not a problem, because the mobile content service doesn't support generating the content in the first place. So, if someone requests a definition from a non-main namespace, we will end up with some nonsense, that will be cached. We might want to disable caching for a definition endpoint in RESTBase

PR #18, opened 2016-04-29
Bring mobile-sections update to change-prop

Pchelolo:
One more piece of restbase update migrations.

This is the last PR needed for complete migration of restbase hacky updates functionality to change-prop

cc @wikimedia/services


d00rman:
LGTM, rebase needed though.

PR #21, opened 2016-05-04
Rule execution: Use geometric progression for the retry delay

d00rman:
The naive approach to retries is not satisfactory. If there is an outage somewhere in the request path (RESTBase, MW, etc), retrying three times 0.5 seconds apart will not help, as these are usually situations where human intervention is needed.

We now use geometric progression to allow sufficient time to pass by so that retries have a chance of succeeding. For that, we introduce a third parameter - `retry_factor` - which is the basis of the progression, and its default value is set to 6. That gives us the following delays if all of the defaults are used: 0.5s, 3s, 18s, 1m48s, 10m48s and 1h4m48s

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -4,6 +4,10 @@ const stringify = require('json-stable-stringify');
 const HyperSwitch = require('hyperswitch');
 const Template = HyperSwitch.Template;
 
+const DEFAULT_RETRY_DELAY = 500;
+const DEFAULT_RETRY_FACTOR = 6;
+const DEFAULT_RETRY_LIMIT = 5;
+
 /**
  * Creates a JS function that verifies property equality
  *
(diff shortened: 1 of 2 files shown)


Pchelolo:
I agree it's a needed and cool feature, but how about allowing simple math expressions and evaluating them? Something similar with what we do in `service-runner` for `ncpu * x` feature. The security wouldn't be a concern as 1. we are the only ones who control what's in the expression 2. We can check if it's a simple expression by a simple whitelist regex.

d00rman:
@Pchelolo what would be the use case? Even simple expressions might get complicated very easily as they would likely want to reference the last delay, which would need to be re-computed (unless we stick in the event message, which would not be ideal).

Users can control all the parameters in the equation `start * base ^ exp` so it's easy to tune it to one's needs.

Pchelolo:
Just a cool feature :) I'm 👍 on this change as is anyway

PR #23, opened 2016-05-04
Added test for startup offset behaviour

Pchelolo:
No code changes, just a test for https://github.com/d00rman/kafka-node/pull/2

So, the tests should fail until that one is settled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ node_js:
   - "5"
 
 env:
-  - KAFKA_HOME=../kafka CXX=g++-4.8
-  - CHANGE_PROP_STOP_DELAY=3000
+  - KAFKA_HOME=../kafka CXX=g++-4.8 CHANGE_PROP_STOP_DELAY=3000
 
 before_install:
 - wget http://www.us.apache.org/dist/kafka/0.8.2.2/kafka_2.10-0.8.2.2.tgz -O kafka.tgz
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
   - "5"
 
 env:
-  - KAFKA_HOME=../kafka CXX=g++-4.8 CHANGE_PROP_STOP_DELAY=3000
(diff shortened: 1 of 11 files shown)


d00rman:
Hm, something strange is going on here. The tests still fail even after publishing the new version of wmf-kafka-node. Master [builds just fine](https://travis-ci.org/wikimedia/change-propagation/builds/127911818) with the latest version, though.

Pchelolo:
Ok, I've figured it out why do the tests pass locally but not in travis - the root cause is that locally you don't delete kafka completely, but just delete topics, while in travis it's a new kafka install. @ottomata's fix for automatic offset reset works for some reason only in the former situation, not the latter. 

Investigation is not over, but I'm already lost..

d00rman:
Ah, yes! So, there's a thing in Kafka where deleting a topic doesn't delete its log. So, when create a topic with the same name, Kafka assigns it a log file with the same name, but doesn't truncate it, so you end up with a new topic, but with the old queue as well. So in local tests we need to delete the log directory as well.

Pchelolo:
@d00rman hm.. my observations do not support that. When I delete the topic and recreate it, the reported offsets start with 0 (-1 actually). If old data was still in there, we'd read it again, but we don't.

Pchelolo:
@d00rman right. But the funny thing is - for some reason this test started to fail for me locally now. I did not change anything, but yesterday it was passing, today it fails. magic. Investigating.

Pchelolo:
I have updated the PR with some new code, but it doesn't work. Just published to not loose. We need updates in the kafka client code for this to work.

Pchelolo:
kafka client - 0, Petr - 1. After merging https://github.com/d00rman/kafka-node/pull/3 this should finally begin to work.

d00rman:
@Pchelolo let's also reference v0.1.3 of wmf-kafka-node in `package.json` given that it is a requirement.

Pchelolo:
@d00rman ye, I didn't do it because it wasn't published yet.

PR #25, opened 2016-05-05
No automatic redirects or retries

d00rman:
The `request` library automatically follows redirects for GET requests which return a 3xx code. This is not a desirable property for change propagation, so disable it. The `preq` library automatically retries failed requests, but since change propagation has its own retry logic, disable that as well.

Additionally, supply the `?redirect=false` query parameter to RESTBase when doing the update requests.

Bug: [T134483](https://phabricator.wikimedia.org/T134483)

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -58,6 +58,8 @@ services:
                         method: get
                         # Don't encode title since it should be already encoded
                         uri: '{{match.meta.uri[1]}}://{{message.meta.domain}}/api/rest_v1/page/summary/{{match.meta.uri[2]}}'
+                        query:
+                          redirect: false
                         headers:
                           cache-control: no-cache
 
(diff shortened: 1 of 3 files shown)


d00rman:
cc @wikimedia/services

d00rman:
Baaah, need to fix the nock mocks in the tests too

config.example.wikimedia.yaml
```diff
@@ -58,6 +58,8 @@ services:
                         method: get
                         # Don't encode title since it should be already encoded
                         uri: '{{match.meta.uri[1]}}://{{message.meta.domain}}/api/rest_v1/page/summary/{{match.meta.uri[2]}}'
+                        query:
+                          redirect: false
```
Pchelolo: Should we add this by default if it's `/api/rest_v1/`?

d00rman: Good point. Will do

d00rman: Ah no we actually can't do that because in prod we do not use the external URI style

Pchelolo: Match on `/v1/`?

d00rman: Too short to match IMHO

d00rman: And too broad. `/v1/` can be contained in so many URIs and is by no means RB-specific.

PR #28, opened 2016-05-11
Concurrency limiting

d00rman:
This PR brings concurrency limiting to event executions. When a message is received, it is sent to the task queue, which either starts the execution right away, or it defers it for later execution, depending on the number of tasks currently in execution. When a task is completed, the list of pending tasks is consulted and, if there are outstanding ones, they are put in the queue.

Note that this PR focuses solely on concurrency limiting, as it is a blocker for re-enabling change propagation in production. A subsequent PR will deal with commit offsets.

Bug: [T134456](https://phabricator.wikimedia.org/T134456)

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -46,15 +46,23 @@ class RuleExecutor {
         return consumer;
     }
 
-    _exec(event, statName) {
-        const rule = this.rule;
-        if (!rule.test(event)) {
-            // no match, drop the message
-            this.log(`debug/${rule.name}`, { msg: 'Dropping event message', event: event });
-            return P.resolve();
+    _test(event) {
+        if (this.rule.test(event)) {
+            return true;
         }
+        // no match, drop the message
+        this.log(`debug/${this.rule.name}`, { msg: 'Dropping event message', event: event });
+        return false;
+    }
+
+    _exec(event, statName, statDelayStartTime) {
+        const rule = this.rule;
 
         this.log(`trace/${rule.name}`, { msg: 'Event message received', event: event });
+
+        // latency from the original event creation time to execution time
+        this.hyper.metrics.endTiming([statName + '_delay'],
+            statDelayStartTime || new Date(event.meta.dt));
 
         const startTime = Date.now();
         const expander = {
@@ -112,9 +120,6 @@ class RuleExecutor {
                         return;
                     }
 
-                    // Latency from the original event creation time to dequeue time
-                    this.hyper.metrics.endTiming([statName + '_delay'], new Date(message.meta.dt));
-
                     if (message.emitter_id !== this._consumerId()) {
                         // Not our business, don't care
                         return;
@@ -125,7 +130,14 @@ class RuleExecutor {
                         return;
                     }
 
-                    return this._exec(message.original_event, statName)
+                    if (!this._test(message.original_event)) {
+                        // doesn't match any more, possibly meaning
+                        // the rule has been changed since we last
+                        // executed it on the message
+                        return;
+                    }
+
+                    return this._exec(message.original_event, statName, new Date(message.meta.dt))
                     .catch((e) => {
                         const retryMessage = this._constructRetryMessage(message.original_event,
                             e, message.retries_left - 1);
@@ -186,14 +198,10 @@ class RuleExecutor {
                     let message;
                     return P.try(() => {
                         message = this._safeParse(msg.value);
-                        if (!message) {
-                            // Don't retry if we can't parse an event, just log.
+                        if (!message || !this._test(message)) {
+                            // no message or no match, we are done here
                             return;
                         }
-
-                        // Latency from the original event creation time to dequeue time
-                        this.hyper.metrics.endTiming([statName + '_delay'],
-                            new Date(message.meta.dt));
 
                         return this._exec(message, statName)
                         .catch((e) => {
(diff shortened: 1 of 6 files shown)


d00rman:
cc @wikimedia/services

d00rman:
This is still a semi-WIP in that it needs some more polishing, but the logic is there.

lib/rule_executor.js
```diff
@@ -174,6 +205,7 @@ class RuleExecutor {
     }
 
     subscribe() {
+        const self = this;
```
gwicke: Is this needed? As far as I can tell, all nested functions are arrow functions, so don't affect `this`.

gwicke:
My concern with the in-memory task queue proposed here is that doesn't seem to be any back-pressure (kafka consumer pausing), so it seems likely that we'd just run out of memory if actual event processing can't keep up.

Previously, we discussed doing something like this: Keep a counter of the currently active (matched & processing) jobs, and pause the kafka consumer whenever the number of concurrently processed tasks exceeds the desired level. Whenever a job is done (success or moved to retry / error), decrement the number of outstanding jobs & un-pause the consumer if concurrency dropped below the threshold.

For the reliable commit, we'd additionally track currently active offsets in an object (add when starting processing, remove when done). Periodically, every second or so, a background task can then look at this object, find the minimum offset being processed (or, if nothing is processed, the client offset), and commit that minus one.

lib/rule_executor.js
```diff
@@ -125,14 +133,33 @@ class RuleExecutor {
...
+                    });
+                    /*return this._exec(message.original_event, statName, new Date(message.meta.dt))
                     .catch((e) => {
                         const retryMessage = this._constructRetryMessage(message.original_event,
                             e, message.retries_left - 1);
                         if (this.rule.shouldRetry(e) && !this._isLimitExceeded(retryMessage)) {
                             return this._retry(retryMessage);
                         }
-                    });
+                    });*/
```
Pchelolo: Why are we just commenting-out this code?

lib/rule_executor.js
```diff
@@ -144,26 +171,30 @@ class RuleExecutor {
...
             Math.pow(spec.retry_factor, spec.retry_limit - retryMessage.retries_left);
         return P.delay(delay)
-        .then(() => this.hyper.post({
-            uri: '/sys/queue/events',
-            body: [ retryMessage ]
-        }));
+        .then(() => {
+            const now = new Date();
+            retryMessage.meta.id = uuid.fromDate(now).toString();
+            retryMessage.meta.dt = now.toISOString();
```
Pchelolo: The `enqueue` endpoint puts these in transparently, why don't we use that?

lib/rule_executor.js
d00rman: Indeed, it is not. Will remove.

lib/rule_executor.js
d00rman: Because the PR is a semi-WIP. That'll go away, for sure.

lib/rule_executor.js
d00rman: Good catch! No need for copy/pasta.

d00rman:
> My concern with the in-memory task queue proposed here is that there doesn't seem to be any back-pressure (kafka consumer pausing). It seems likely that we'd just run out of memory if actual event processing can't keep up with dequeue rate.

Yup, that's my concern too, but didn't manage to get that detail in the PR yesterday. Will do so today.

> For the reliable commit, we'd additionally track currently active offsets in an object (add when starting processing, remove when done). Periodically, every second or so, a background task can then look at this object, find the minimum offset being processed (or, if nothing is processed, the client offset), and commit that minus one.

Commit the right offset needs more work in the Kafka client, so IMHO it's better left for a subsequent PR. While committing **any** offset is definitely a bug, it's not a bug that precludes us from reactivating Change Prop in prod and having parallel updates from both RESTBase and Change Prop. By the time we are sure all is working as promised, commit offsets will be fixed.

lib/rule_executor.js
d00rman: Euh, actually it is because we encapsulate the arrow functions inside Objects.

d00rman:
Consumer pausing / resume has been implemented as well.

I have tested this extensively in MW-Vagrant by simulating an influx of messages and the mechanism works correctly. Next steps:
- set up conservative concurrency limits
- test in beta
- move testing to prod

d00rman:
Published as [v0.2.4](https://github.com/wikimedia/change-propagation/tree/v0.2.4)

lib/rule_executor.js
```diff
@@ -89,12 +99,14 @@ class RuleExecutor {
...
         return false;
     }
+
     /**
      * Set's up a consumer a retry queue
      *
      * @private
      */
     _setUpRetryTopic() {
+        var self = this;
```
Pchelolo: `var`? `self`?

gwicke:
@d00rman, I wondered why the task queue is needed at all, so did a bit of digging. This is my understanding: Calling `pause()` on the consumer only affects fetching, but not event emission. As events are fetched in chunks, this means that we'd potentially exceed the desired concurrency limit by starting requests for a full chunk. The task queue avoids this by buffering surplus requests from the chunk until the limit is no longer exceeded.

Is this about right? If so, I would like to add a comment explaining this somewhere.

d00rman:
@gwicke exactly. Just having counters or offset IDs is not enough because of this pre-fetch. I'll add a comment in the header of the task queue file.

gwicke:
I now actually wondering if we could do this queuing in the driver, for the benefit of other users. But, that's a separate conversation.

@d00rman, thanks for the clarification!

PR #29, opened 2016-05-16
Commit only offsets that were already processed

Pchelolo:
By default the `HighLevelConsumer` commits offsets of the most recent fetched messages. This is obviously incorrect as we want to commit the offset only after a task execution has been finished (either rejected, or successfully finished, or failed). However, if the service crashes while the task was executed, we don't want to commit the offset. 

As we have a `TaskQueue` now, we can track task execution, and commit proper offset when the task is finished (which is the minimal offset in the queue for the tasks consumerGroup, topic and partition).

As the commit itself is a heavy operation, we accumulate pending commits for some interval and commit all at once.

Also, introduced a simple `Task` class as an immutable (as immutable as it could be in JS) object to hold task properties and a couple of utility getters and functions.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -101,8 +101,10 @@ class KafkaFactory {
      * @param {string} topic topic to commit offset for
      * @param {Object} offsets offset values to commit for partitions. Object keys
      *                         represent partitions, values - offsets.
+     * @param {Function} shouldResume a no-arg callback checking whether the consumer
+     *                                should be resumed after offsets have been committed
      */
-    static commit(consumer, topic, offsets) {
+    static commit(consumer, topic, offsets, shouldResume) {
         function setOffsets(values) {
             Object.keys(values).forEach((partition) => {
                 consumer.setOffset(topic, partition, values[partition]);
@@ -119,7 +121,9 @@ class KafkaFactory {
         return consumer.commitAsync(true)
         .then(() => {
             setOffsets(oldOffsets);
-            consumer.resume();
+            if (shouldResume()) {
+                consumer.resume();
+            }
         });
     }
 }
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -93,6 +93,35 @@ class KafkaFactory {
     get produceDC() {
         return this.kafkaConf.dc_name || this.kafkaConf.produce_dc || 'datacenter1';
     }
+
+    /**
+     * Commit a specific offset.
+     *
+     * @param {kafka.HighLevelConsumer} consumer a consumer to commit offset for
+     * @param {string} topic topic to commit offset for
+     * @param {Object} offsets offset values to commit for partitions. Object keys
+     *                         represent partitions, values - offsets.
+     */
+    static commit(consumer, topic, offsets) {
+        function setOffsets(values) {
+            Object.keys(values).forEach((partition) => {
+                consumer.setOffset(topic, partition, values[partition]);
+            });
+        }
+
+        const oldOffsets = {};
+        consumer.pause();
+        consumer.topicPayloads.filter((p) => p.topic === topic)
+        .forEach((p) => {
+            oldOffsets[p.partition] = p.offset;
+        });
+        setOffsets(offsets);
+        return consumer.commitAsync(true)
+        .then(() => {
+            setOffsets(oldOffsets);
+            consumer.resume();
+        });
+    }
 }
 
 module.exports = KafkaFactory;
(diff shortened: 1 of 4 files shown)


lib/kafka_factory.js
```diff
@@ -93,6 +93,35 @@ class KafkaFactory {
...
+        consumer.pause();
+        consumer.topicPayloads.filter((p) => p.topic === topic)
+        .forEach((p) => {
+            oldOffsets[p.partition] = p.offset;
+        });
+        setOffsets(offsets);
+        return consumer.commitAsync(true)
+        .then(() => {
+            setOffsets(oldOffsets);
+            consumer.resume();
```
d00rman: There's a race condition here. Between the execution of L119 and this one the consumer might have been paused by `TaskQueue._defer()` and should not be resumed in that case. A simple instance variable flag can fix this.

lib/task_queue.js
```diff
@@ -3,10 +3,12 @@
 
 const EventEmitter = require('events');
 const P = require('bluebird');
+const KafkaFactory = require('./kafka_factory');
 
 
 /** @const {number} the dafault queue size */
 const DEFAULT_SIZE = 100;
+const COMMIT_TIMEOUT = 100;
```
d00rman: Nit: add a comment before the const.

lib/task_queue.js
```diff
@@ -19,17 +21,22 @@ class TaskQueue extends EventEmitter {
...
         this._max_size = options.size || DEFAULT_SIZE;
+        this._commit_timeout = options.commitTimeout || COMMIT_TIMEOUT;
         this._queue = [];
         this._waiting = [];
         this._consumers = [];
+        this._commitScheduled = false;
+        this._pendingCommits = new Map();
+
         this.on('task_completed', this._deferred.bind(this));
+        this.on('task_completed', this._commitOffsets.bind(this));
```
d00rman: IMHO, the code would be more naturally read if `this._commitOffsets` were called from `this._deferred()` given that it is the one resuming the consumers.

lib/task_queue.js
```diff
@@ -98,6 +156,29 @@ class TaskQueue extends EventEmitter {
...
+     *
+     * @private
+     */
+    _commitOffsets() {
+        if (this._commitScheduled) {
+            return;
+        }
+
+        setTimeout(() => {
+            for (let entry of this._pendingCommits.entries()) {
```
d00rman: Wouldn't using `Map.forEach()` make this code more readable?

lib/task_queue.js
```diff
@@ -71,22 +77,74 @@ class TaskQueue extends EventEmitter {
...
     }
 
+    /**
+     * Finds the lowest offset that is safe to commit after some task finished.
+     *
+     * @param {Task} task just finished task
+     * @private
+     */
+    _updatePendingCommits(task) {
+        const current = this._pendingCommits.get(task.consumer)
```
d00rman: I think it pays of here to have `const consumer = this._pendingCommits.get(task.consumer)` and then use it throughout the method as it's used quite a few times.

Pchelolo:
@d00rman Updated, addressed your comments

lib/task_queue.js
```diff
@@ -71,33 +90,112 @@ class TaskQueue extends EventEmitter {
...
+    _commitOffsets() {
+        if (this._commitScheduled) {
+            return;
+        }
+
+        setTimeout(() => {
+            this._pendingCommits.forEach((topicsOffsets, consumer) => {
+                Object.keys(topicsOffsets).forEach((topic) => {
+                    KafkaFactory.commit(consumer, topic,
+                        topicsOffsets[topic], () => !this._isPaused);
```
d00rman: Did you mean(TM) `!this.is_paused()` ?

Pchelolo: No, I meant what I've written. 

Idea: isPaused is a flag in `TaskQueue` that's set to true if taskQueue is paused, and reset to false when it's resumed. The callback function I'm providing is called `shouldResume`, so, when it returns `false` (when queue is paused) the commit method doesn't resume the consumer.

Why it's a function you would ask, and not a bool flag. Because commit is async, so from the time when the flag is captured to the time when it's used, it's value might have already been changed, as some new message might have come and the queue might have been filled up.

lib/task_queue.js
```diff
@@ -19,16 +33,22 @@ class TaskQueue extends EventEmitter {
...
-     *
+     * @param {number} [options.commitTimeout=COMMIT_TIMEOUT] the maximum delay to commit
+     *                                                        the offsets
      * @constructor
      */
     constructor(options) {
         super();
         options = options || {};
         this._max_size = options.size || DEFAULT_SIZE;
+        this._commit_timeout = options.commitTimeout || COMMIT_TIMEOUT;
```
d00rman: Let's do it vice-versa: call the instance var `_commitTimeout` and the option `commit_timeout` as we use snake_case for options and camelCase for the instance var.

PR #32, opened 2016-05-16
Supply triggered_by fields and provide x-triggered-by headers

Pchelolo:
We've decided to have an analogue of the XFF header in events, to prevent loops and for debugging purposes. This PR makes change-prop fill those properties.

Depends on https://gerrit.wikimedia.org/r/#/c/288918/

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -191,9 +191,9 @@ class RuleExecutor {
                 request_id: event.meta.request_id,
                 id: undefined, // will be filled later
                 dt: undefined, // will be filled later
-                domain: event.meta.domain,
-                triggered_by: utils.triggeredBy(retryEvent || event)
+                domain: event.meta.domain
             },
+            triggered_by: utils.triggeredBy(retryEvent || event),
             emitter_id: this._consumerId(),
             retries_left: retriesLeft === undefined ? this.rule.spec.retry_limit : retriesLeft,
             original_event: event,
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -4,6 +4,7 @@ const P = require('bluebird');
 const uuid = require('cassandra-uuid').TimeUuid;
 const TopicsNotExistError = require('wmf-kafka-node/lib/errors').TopicsNotExistError;
 const Task = require('./task');
+const utils = require('./utils');
 
 /**
  * A rule executor managing matching and execution of a single rule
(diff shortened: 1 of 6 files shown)


lib/rule_executor.js
```diff
@@ -178,7 +182,8 @@ class RuleExecutor {
                 request_id: event.meta.request_id,
                 id: undefined, // will be filled later
                 dt: undefined, // will be filled later
-                domain: event.meta.domain
+                domain: event.meta.domain,
+                triggeredBy: triggeredBy
```
gwicke: Minor: Should we stick with snake_case for consistency?

Pchelolo: @gwicke Ok, will change. But pls don't review any more until the WIP prefix disappears. It's a lot of work still, I'm made a PR before I've realised it's only a part of the solution.

gwicke: Yepyep, just noticed this while catching up on what's going on ;)

Pchelolo:
@wikimedia/services This is not a WIP any more, it's ready for review.

test/feature/static_rules.js
```diff
@@ -155,13 +165,27 @@ describe('Basic rule management', function() {
             reqheaders: {
                 test_header_name: 'test_header_value',
                 'content-type': 'application/json',
-                'x-request-id': common.SAMPLE_REQUEST_ID
+                'x-request-id': common.SAMPLE_REQUEST_ID,
```
d00rman: Let's not :P

test/feature/update_rules.js
```diff
@@ -103,11 +105,13 @@ describe('RESTBase update rules', function() {
...
         const mwAPI = nock('https://en.wikipedia.org', {
             reqheaders: {
-                'cache-control': 'no-cache'
+                'cache-control': 'no-cache',
+                'x-triggered-by': 'resource_change:https://en.wikipedia.org/api/rest_v1/page/html/Main%20Page'
             }
         })
         .get('/api/rest_v1/page/mobile-sections/Main%20Page')
         .query({ redirect: false })
+
```
d00rman: Not needed.

test/feature/static_rules.js
Pchelolo: Let's not what? :)

Pchelolo: Let's not what? :)

lib/utils.js
```diff
@@ -0,0 +1,17 @@
...
+const utils = {};
+
+/**
+ * Computes the x-triggered-by header
+ *
+ * @param {Object} event the event
+ * @returns {string}
+ */
+utils.triggeredBy = (event) => {
+    return event.triggered_by ?
```
d00rman: I'd prefer to have something that consolidates the addition of the new part (instead of copy/pasting it on two lines):

```js
const prevTrigger = event.triggered_by || '';
if (prevTrigger) { prevTrigger += ','; }
return prevTrigger + event.meta.topic + ':' + event.meta.uri;
```

test/feature/static_rules.js
d00rman: Let's not add the ending comma :P

Pchelolo: It's not ending, there's one more line in this object (line 63)

d00rman: Hm, how did I miss that?

PR #33, opened 2016-05-17
Various stability improvements

Pchelolo:
Today we've had some weird retry loop in change propagation again. While the root cause of the original issue is still not clear, while investigating it we've found several areas to improve the service:

1. In case there's an exception in the _test method, it crashes the service - fixed by wrapping in try-catch
2. The `x-request-id` header is not passed to RESTBase
3. There's a rase condition on consumer resuming. Fixed by adding an `isCommitting` flag
4. Increased commit interval to 500ms

cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -74,7 +74,7 @@ function _compileMatch(obj, result, name, fieldName) {
             result[fieldName] = obj;
             return `${name} === ${obj}`;
         }
-        if (obj[0] !== '/' || obj[obj.length - 1] !== '/') {
+        if (!/^\/.+\/[gimuy]{0,5}$/.test(obj)) {
             // not a regex, quote the string
             result[fieldName] = `'${obj}'`;
             return `${name} === '${obj}'`;
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -74,7 +74,7 @@ function _compileMatch(obj, result, name, fieldName) {
             result[fieldName] = obj;
             return `${name} === ${obj}`;
         }
-        if (obj[0] !== '/' && obj[obj.length - 1] !== '/') {
+        if (obj[0] !== '/' || obj[obj.length - 1] !== '/') {
             // not a regex, quote the string
             result[fieldName] = `'${obj}'`;
             return `${name} === '${obj}'`;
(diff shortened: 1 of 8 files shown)


lib/rule.js
```diff
@@ -74,7 +74,7 @@ function _compileMatch(obj, result, name, fieldName) {
             result[fieldName] = obj;
             return `${name} === ${obj}`;
         }
-        if (obj[0] !== '/' && obj[obj.length - 1] !== '/') {
+        if (obj[0] !== '/' || obj[obj.length - 1] !== '/') {
```
d00rman: Heh, this neat trick allowed for flags to be passed to regexes as well, e.g. `/searchstring/i` but it complicates our rules with URIs.

Pchelolo: Let's make it more precise then, smth like `^\/.+\/[gimuy]{0,5}$/`?

lib/task_queue.js
```diff
@@ -20,7 +20,7 @@ const DEFAULT_SIZE = 100;
  * @const
  * @type {number}
  */
-const COMMIT_TIMEOUT = 100;
+const DEFAULT_COMMIT_TIMEOUT = 500;
```
d00rman: Actually, a more appropriate name would be `DEFAULT_COMMIT_INTERVAL` since that's what this really is. Likewise, it'd be good to rename `commitTimeout` into `commitInterval`

Pchelolo: @d00rman eh, not really. If we have a really low volume of messages, the timeout will not be set right away after the previous commit, so an actual interval could be larger then the COMMIT_TIMEOUT

lib/rule.js
d00rman: That's slower than looking at two chars, so let's keep it as-is for now until we see that it is really needed.

lib/task_queue.js
d00rman: Even in light of low message volumes, to me `COMMIT_INTERVAL` says *at most 1 commit will happen in the next `COMMIT_INTERVAL` milliseconds*. It's a nit, really, but I tend to associate the word *timeout* with bad stuff :)

d00rman:
Published as [v0.3.1](https://github.com/wikimedia/change-propagation/tree/v0.3.1)

PR #34, opened 2016-05-20
Support named groups in regex

Pchelolo:
Just a bit of syntactic sugar - when the match elements are taken from a regex, we need to reference them via indexes, which is not really convenient. Instead, support naming capture groups with the following syntax: `(?<name>group_regex)` and reference them by name.

Named groups cannot be mixed with unnamed groups.

cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -66,6 +66,12 @@ function _compileNamedRegex(obj, result, name, fieldName) {
         captureNames.push(name);
         return '(';
     });
+    const numGroups = (new RegExp(normalRegex.toString() + '|')).exec('').length - 1;
+
+    if (captureNames.length && captureNames.length !== numGroups) {
+        throw new Error('Invalid match regex. ' +
+            'Mixing named and unnamed capture groups are not supported. Regex: ' + obj);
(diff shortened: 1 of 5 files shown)


Pchelolo:
The tests will fail until this one is rebased against https://github.com/wikimedia/change-propagation/pull/32

lib/rule.js
```diff
@@ -61,6 +60,27 @@ function _compileArrayMatch(obj, result, name) {
     return `Array.isArray(${name})` + (itemsCheck.length ? (' && ' + itemsCheck) : '');
 }
 
+function _compileNamedRegex(obj, result, name, fieldName) {
+    const captureNames = [];
+    const normalRegex = obj.replace(/\(:<(\w+)>/g, (_, name) => {
```
d00rman: The usual syntax for named groups is `(?<name>`, so it'd be better/easier to use `?` instead of `:`

d00rman:
> Named groups cannot be mixed with unnamed groups.

The code doesn't ensure that at the moment. Let's make it so. And also add a test for this case.

PR #37, opened 2016-05-25
Added rules to partially replace RESTBaseUpdateJobs

Pchelolo:
These rules are doing exactly the same thing as RBUpdateJobs used to do. They replace everything except transfusions tracking and updating. 

This was tested via vagrant for almost the whole day, and it seems that everything works just fine.

Cannot enable them until https://gerrit.wikimedia.org/r/#/c/290687/ is deployed in production.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -119,3 +119,79 @@ services:
                         body:
                           - meta:
                               uri: '//{{message.meta.domain}}/api/rest_v1/{{match.meta.uri.title}}'
+
+                    # RESTBase update jobs
+                    mw_purge:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/^(?<proto>https?):\/\/[^\/]+\/wiki\/(?<title>.+)$/'
+                        tags:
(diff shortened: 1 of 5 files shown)


d00rman:
Veeeery nice! :+1:

Pchelolo:
@d00rman hehe, when you spend a whole day on one PR it must be good ;)

Pchelolo:
@d00rman Change in EventBus was deployed, so we can merge this one. Once it's merged I'll prepare changes for puppet and vagrant.

d00rman:
> @d00rman Change in EventBus was deployed, so we can merge this one. Once it's merged I'll prepare changes for puppet and vagrant.

On `1.28.0-wmf.3` which isn't on all projects yet (bound to happen late this evening), so we can't deploy it yet. Tomorrow is friday, which leaves monday.

Pchelolo:
@d00rman sure. Just merging it will be a no-op, and I will create puppet and vagrant patches for Monday.

d00rman:
Sounds like a plan!

PR #38, opened 2016-05-26
File transclusions processing

Pchelolo:
There're two types of dependency updates left in the update jobs:
 - File transclusions
 - Other transclusions

The former is less high-volume (at least it's unlikely that one file would be used in tens of thousands of pages), so let's start examining the continuation module by porting only file transclusions.  Right now it's a direct port of what RBUpdateJob is doing, but there's a couple of enhancements we might wanna do later - decide what to do with redirect pages for example.

In order for this to work we need to set up a new topic for transcludes continuation.

Tested with vagrant, it seems to work, but proper labs testing is required.
Also, it's not clear what would happen for files on 'commons' used in the wikis, I obviously cannot test it in vagrant.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -34,6 +34,8 @@ services:
                         uri: 'https://{{message.meta.domain}}/w/api.php'
                         headers:
                           host: '{{message.meta.domain}}'
+                        body:
+                          formatversion: 2
           /sys/queue:
             x-modules:
               - path: sys/kafka.js
@@ -233,3 +235,5 @@ services:
                         uri: 'https://{{message.meta.domain}}/api/rest_v1/page/html/{{match.meta.uri.title}}'
                         headers:
                           cache-control: no-cache
+                        query:
+                          redirect: false
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -27,7 +27,7 @@ services:
                   port: 4321
           /sys/links:
             x-modules:
-              - path: sys/backlinks.js
+              - path: sys/dep_updates.js
                 options:
                   templates:
                     mw_api:
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -221,3 +221,15 @@ services:
                         method: 'post'
                         uri: '/sys/links/transcludes/{{message.page_title}}'
                         body: '{{globals.message}}'
+
+                    transclusion_updates:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/https:\/\/[^\/]+\/wiki\/(?<title>.+)/'
+                        tags: [ 'change-prop', 'transcludes' ]
+                      exec:
+                        method: get
+                        uri: 'https://{{message.meta.domain}}/api/rest_v1/page/html/{{match.meta.uri.title}}'
+                        headers:
+                          cache-control: no-cache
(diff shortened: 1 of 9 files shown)


config.example.wikimedia.yaml
```diff
@@ -214,3 +214,22 @@ services:
...
+
+                    transclusion_updates:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/https:\/\/[^\/]+\/wiki\/(?<title>.+)/'
+                        tags: [ 'change-prop', 'transcludes' ]
+                      exec:
+                        method: get
+                        uri: 'https://{{message.meta.domain}}/api/rest_v1/page/html/{{match.meta.uri.title}}'
```
d00rman: `redirect: false` also needs to be here.

config.example.wikimedia.yaml
```diff
@@ -214,3 +214,22 @@ services:
...
+                        method: 'post'
+                        uri: '/sys/links/transcludes/{{message.page_title}}'
+                        body: '{{globals.message}}'
+
+                    transclusion_updates:
+                      topic: resource_change
+                      match:
+                        meta:
+                          uri: '/https:\/\/[^\/]+\/wiki\/(?<title>.+)/'
+                        tags: [ 'change-prop', 'transcludes' ]
```
d00rman: This is an `OR` or an `AND` ?

d00rman: Nm, it's an `AND`

sys/backlinks.js
```diff
@@ -21,6 +22,74 @@ class BackLinksProcessor {
...
+        extractResults: (res) => res.body.query.backlinks
+    };
+}
+
+function createImageUsageTemplate(options) {
+    return {
+        template: new Template(extend(true, {}, options.templates.mw_api, {
+            method: 'post',
+            body: {
+                format: 'json',
```
d00rman: Let's use `formatversion: 2` as it's the upcoming default (perhaps might be a good idea to put it directly in the `mw_api` template)

d00rman:
Looking very good!

As for commons and other interwiki usages, we'll have to use `prop=iwlinks`. For example, for https://commons.wikimedia.org/wiki/File:Official_portrait_of_Barack_Obama.jpg :
- [imageusage](https://commons.wikimedia.org/w/api.php?action=query&format=json&formatversion=2&list=imageusage&iutitle=File:Official_portrait_of_Barack_Obama.jpg&iulimit=500) gives you only the pages on commons that reference it
- [iwlinks](https://commons.wikimedia.org/w/api.php?action=query&format=json&formatversion=2&prop=iwlinks&titles=File:Official_portrait_of_Barack_Obama.jpg) retrieves all the image's usages on all other wikis (but not commons)

Hence, these two are complimentary. I think we can leave `iwlinks` for a second step, since *(a)* RBUpdateJobs wasn't handling it anyway; and *(b)* `iwlinks` will have to be implemented for most updates and transclusions anyway (hint: wikidata).

config.example.wikimedia.yaml
```diff
@@ -214,3 +216,24 @@ services:
                             if-unmodified-since: '{{date(message.meta.dt)}}'
                           query:
                             redirect: false
+
+                    transclusion_update:
+                      topic: mediawiki.revision_create
+                      exec:
+                        method: 'post'
+                        uri: '/sys/links/transcludes/{{message.page_title}}'
+                        body: '{{globals.message}}'
```
d00rman: What is `globals.message` ?

Pchelolo: This will get resolved to the complete message. The problem with doing just `{{message}}` is that it's ambiguous for the template processor, whether it's a shortcut for `{{request.body.message}}` or whether it's a field from the global namespace. This is a workaround for the syntactic ambiguity.

PR #39, opened 2016-05-27
Added test for redirects, added status monitoring

Pchelolo:
Several small improvements:
1. As we do not follow redirects, and must use normalised titles already, receiving a 301 response is a bug and requires looking into - thus log a warning.
2. There was a bug in `hyperswitch` - it stripped out the `followRedirect` property we set. Added a test for this.
3. There's a support in service-runner to log the status message from a worker when it dies - utilise this feature to get more information on debugging.

Depends on https://github.com/wikimedia/hyperswitch/pull/44

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -37,7 +37,7 @@ class RuleExecutor {
                     rule: { name: ruleName, topic },
                     msg: `Listening to ${topicList[0]}`
                 });
-                this._notifySubscribed(topic)
+                this._notifySubscribed(topic);
             } else {
                 this.log(`info/subscription/${ruleName}`, {
                     rule: { name: ruleName, topic },
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -85,7 +85,18 @@ class RuleExecutor {
                 'x-request-id': origEvent.meta.request_id,
                 'x-triggered-by': utils.triggeredBy(retryEvent || origEvent)
             });
-            return this.hyper.request(request);
+            return this.hyper.request(request)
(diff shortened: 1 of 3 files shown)


lib/rule_executor.js
```diff
@@ -85,7 +86,18 @@ class RuleExecutor {
                 'x-request-id': origEvent.meta.request_id,
                 'x-triggered-by': utils.triggeredBy(retryEvent || origEvent)
             });
-            return this.hyper.request(request);
+            return this.hyper.request(request)
+            .tap((res) => {
+                if (res.status === 301) {
```
d00rman: Hm, this behaviour is true for RB, but the policy should not be applied to everything.

Pchelolo: For RB we know there should be no redirects, but what about other stuff then - if we don't follow a redirect we might not actually update something, if we follow - we might get into a loop. 

As RB is currently the only one we're updating - may be let's leave it in, and revisit the policy as we add more update targets? In general, I think the policy should be 'not to use 301 redirects ever'

d00rman: > For RB we know there should be no redirects, but what about other stuff then - if we don't follow a redirect we might not actually update something, if we follow - we might get into a loop.

Which is why we might want to have that a property of the rule - whether to follow redirects or not. Rule creators will know the answer to that and whether it's safe to do so.

> As RB is currently the only one we're updating - may be let's leave it in, and revisit the policy as we add more update targets? In general, I think the policy should be 'not to use 301 redirects ever'

Yup, agreed. 301s just shouldn't happen.

PR #42, opened 2016-05-30
User-Agent is not propagated by change-prop

Pchelolo:
After https://github.com/wikimedia/hyperswitch/pull/41 change-prop doesn't forward a `User-Agent` header to services. Fix it. 

PS: this will require a corresponding change in puppet, this PR is effectively a no-op in prod.

cc @wikimedia/services


Pchelolo:
puppet change: https://gerrit.wikimedia.org/r/291784

PR #45, opened 2016-06-03
Don't decode the results by default

Pchelolo:
Instead of the https://gerrit.wikimedia.org/r/#/c/292366/

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -199,6 +199,9 @@ class RuleExecutor {
                         this._exec.bind(this, message.original_event,
                             statName, new Date(message.meta.dt), message),
                         (e) => {
+                            if (Buffer.isBuffer(e.body)) {
+                                e.body = JSON.parse(e.body.toString());
+                            }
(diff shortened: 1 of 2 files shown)


lib/rule_executor.js
```diff
@@ -199,6 +199,9 @@ class RuleExecutor {
                         this._exec.bind(this, message.original_event,
                             statName, new Date(message.meta.dt), message),
                         (e) => {
+                            if (Buffer.isBuffer(e.body)) {
+                                e.body = JSON.parse(e.body.toString());
```
d00rman: Let's do the parsing in a `try/catch` block, and if it fails, report the string only.

lib/rule_executor.js
```diff
@@ -309,6 +312,9 @@ class RuleExecutor {
                             msg,
                             this._exec.bind(this, message, statName),
                             (e) => {
+                                if (Buffer.isBuffer(e.body)) {
+                                    e.body = JSON.parse(e.body.toString());
```
d00rman: Idem.

PR #48, opened 2016-06-06
Rerender summary when wikidata item is updated

Pchelolo:
When a wikidata page is changed, the item description might have been changed too - thus rerender summary as we wanna be able to include the description into the summary and purge it.

Related PR: https://github.com/wikimedia/restbase/pull/625
Bug: https://phabricator.wikimedia.org/T128894

cc @wikimedia/services

### DIFF
--- a/sys/dep_updates.js
+++ b/sys/dep_updates.js
@@ -88,6 +88,7 @@ function createWikidataTemplate(options) {
             }
         })),
         resourceChangeTag: 'wikidata',
+        shouldProcess: (res) => { return res && res.body && !res.body.success; },
         extractResults: (res) => {
             const siteLinks = res.body.entities[Object.keys(res.body.entities)[0]].sitelinks;
             return Object.keys(siteLinks).map((siteId) => {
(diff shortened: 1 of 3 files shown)


gwicke:
:+1:

sys/dep_updates.js
```diff
@@ -75,13 +75,41 @@ function createTranscludeInTemplate(options) {
...
+                format: 'json',
+                action: 'wbgetentities',
+                ids: '{{message.page_title}}',
+                props: 'sitelinks/urls',
+                normalize: true
+            }
+        })),
+        resourceChangeTag: 'wikidata',
+        extractResults: (res) => {
+            const siteLinks = res.body.entities[Object.keys(res.body.entities)[0]].sitelinks;
```
d00rman: It would be good to protect ourselves from stale edits (quick edits followed by page/revision deletions) and check that we actually have the info we need.

d00rman:
One minor comment, otherwise LGTM.

sys/dep_updates.js
Pchelolo: Great idea. I actually need to protect other transclusions as well, will update others in a separate PR

PR #50, opened 2016-06-07
Test stability improvements

Pchelolo:
There was several issues with test stability:
 -  If one test failed for some reason, it could post an unexpected retry message that's processed by another test and fail it too. Separated the tests better by introducing a randomised field to the test events so that events from different tests don't mix up.
 -  When we call `service-runner.stop()` we only close the HTTP server created by `huperswitch`. But for change-propagation that's not enough - we also need to close kafka consumers, or some events might be processed twice, thus the tests were failing. 
I can't say I like the solution with an artificial `close` endpoint, but that's the simplest one. Alternative options: either make a module emit `server_close` event when the server is closed or introduce a concept of a module 'destructor'. This is not really important in production since we're not using `service-runner.close` there.
 -  Some of the tests were verifying that something is NOT done, and a nock endpoint is not called. Implementing a better way to verify that by comparing the `pendingMocks` with an expected number of pending mocks and used it in tests

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -19,6 +19,16 @@ services:
                     user-agent: true
         title: The Change Propagation root
         paths:
+          # Test-only endpoint for a graceful service shutdown
+          # It's needed in the
+          /v1/close:
+            post:
+              x-request-handler:
+                - do_close:
+                    request:
+                      method: post
+                      uri: /sys/queue/close
+
           /sys/purge:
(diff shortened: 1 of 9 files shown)


Pchelolo:
I've got a better way to do it, will update soon, please don't review yet.

Pchelolo:
Will be updated when we settle on https://github.com/wikimedia/hyperswitch/pull/47

PR #51, opened 2016-06-09
Do not decrement the offset to commit

d00rman:
According to the Kafka docs:

> Note: The committed offset should always be the offset of the next
> message that your application will read.

Hence, decrementing the offset actually means that the message being
currently worked on will be processed twice in case another Change Prop
worker obtains ownership of the partition.


Pchelolo:
👍

PR #53, opened 2016-06-10
Don't pause-resume while rebalancing

Pchelolo:
It seems like pausing/resuming consumers during a rebalance is not a good idea.

cc @wikimedia/services

### DIFF
--- a/lib/task_queue.js
+++ b/lib/task_queue.js
@@ -64,6 +64,11 @@ class TaskQueue extends EventEmitter {
         if (this._consumers.some((item) => item === consumer)) {
             return false;
         }
+        consumer.on('rebalanced', () => {
+            if (!this._isPaused && consumer.paused) {
+                consumer.resume();
+            }
+        });
         this._consumers.push(consumer);
         return true;
     }
--- a/lib/task_queue.js
+++ b/lib/task_queue.js
@@ -164,7 +164,10 @@ class TaskQueue extends EventEmitter {
 
         // enqueue any outstanding tasks
         while (this._queue.length < this._max_size && this._waiting.length > 0) {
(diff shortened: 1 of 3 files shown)


d00rman:
Definitely we are onto something here. I think there's one more piece missing. The task queue doesn't account for the fact that from the time an event entered the queue (either executed directly or put on hold) till the time it's acted upon (both in the sense of execution and offset commit), the consumer might not be the owner of a partition any more. To resolve this, the first thing that pops to mind is to track the message-consumer relationship and, on `rebalanced`, check whether the consumer is still the owner of the given topic/partition. If that's not the case, the message should be discarded. `consumer.topicPayloads` contains the list of topics/partitions the consumer is subscribed to, and given that we have one consumer per rule, this should as easy as checking the length of that list.

lib/task_queue.js
```diff
@@ -164,7 +164,10 @@ class TaskQueue extends EventEmitter {
 
         // enqueue any outstanding tasks
         while (this._queue.length < this._max_size && this._waiting.length > 0) {
-            this._start(this._waiting.shift());
+            const task = this._waiting.shift();
+            if (task.consumer.topicPayloads.some((p) => p.topic === task.topic)) {
```
d00rman: Perhaps to optimise memory, it'd be better to evict them from the waiting queue right on `rebalanced` (but still perform this check for safety) ?

gwicke:
> The task queue doesn't account for the fact that from the time an event entered the queue (either executed directly or put on hold) till the time it's acted upon (both in the sense of execution and offset commit), the consumer might not be the owner of a partition any more. 

Is the main concern about processing those events twice, or are there other errors / conflicts to consider within ChangeProp?

PR #57, opened 2016-06-15
Encode URI of the transclusion resource

Pchelolo:
We need to encode the title for the `resource_change` events which we produce from the transclusions updates rule.

cc @wikimedia/services


d00rman:
This needs a rebase, mediawiki-title is already in master

PR #58, opened 2016-06-16
Epic: Bug fix: Do not allow empty messages to be sent to Kafka

d00rman:
For reasons not yet clear at this point, the Promise sending messages to
Kafka never gets fulfilled when an empty array is given to it, which
causes the task queue to accumulate messages and never discard them,
effectively halting the processing for all of the rules that the worker
is managing.

This patch adds a check to the producer end to reject empty arrays (as
it should, regardless of the underlying cause), and intructs the
transclusion updates not to send resource_change events when no titles
are to be re-rendered.


d00rman:
cc @wikimedia/services

PR #59, opened 2016-06-17
Log all errors that go to error topic

Pchelolo:
Sometimes we see some elevated rates of errors going to the error topic, but they're not seen in the logs. These are the errors which do no match the retry pattern. While we're stabilising CP it would be useful to see these errors in the logs.

cc @wikimedia/services


Pchelolo:
Actually we don't need to do this, it's better to look at the error topic itself.

PR #60, opened 2016-06-22
Added ability to supply switch statements to the rule

Pchelolo:
Sometimes we need to be able to choose an action in one rule out of several options. This change adds the `switch` keyword support to the rule engine. Now `test` method returns the index of the matched option, which should be provided to the `expand` and `exec` methods afterwards.

I don't like the syntax too much, but that's the only way to preserve backwards compatibility of the syntax. Perfectly, I'd like to introduce a required `handler` keyword that would wrap the `match/exec` section or an array of such sections with the switch semantics. But that's backwards incompatible, so we can bike shed on it later

Also, this lacks unit tests, I will add them later. Also, the documentation needs to be updated, but that's also later when we settle on the syntax.

cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -132,40 +132,57 @@ class Rule {
         this.spec.retry_on = this.spec.retry_on || {
             status: [ '50x' ]
         };
-        this.shouldRetry = _compileRetryCondition(this.spec.retry_on);
-        this.exec = this._processExec(this.spec.exec);
-        this._match = this._processMatch(this.spec.match);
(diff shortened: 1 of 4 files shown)


d00rman:
I'm confused wrt `switch`; I don't think it's needed at all. We should just use a straight array. If the rule is an object, just convert it into a one-elem array.

Pchelolo:
@d00rman you have some per-rule global properties: topic, retry delays, retry conditions.

Pchelolo:
Also temporarily commented-out the JSCS tests since something got broken in the dependencies. We will uncomment that when the issue's fixed.

PR #62, opened 2016-06-23
Support ignoring errors

Pchelolo:
We need to support ignoring certain types of errors, for example 412 that means the content has already been rerendered before CP got to it.

cc @wikimedia/services


gwicke:
:+1:

Pchelolo:
Puppet patch with the config changes: https://gerrit.wikimedia.org/r/#/c/295680/

PR #67, opened 2016-06-29
Load rules from files instead of directly specifying them in the config

Pchelolo:
Follow-up from the offsite discussion

cc @wikimedia/services


rules/restbase_deps.yaml
```diff
@@ -0,0 +1,79 @@
+summary_rerender:
```
d00rman: Could collapse summary_rerender and definition_rerender now that we have cases.

sys/kafka.js
```diff
@@ -26,7 +27,16 @@ class Kafka {
             produce_dc: options.produce_dc,
             dc_name: options.dc_name
         });
-        this.staticRules = options.templates || {};
+        this.staticRules = options.rules
```
d00rman: To keep ourselves sane, this change should be backwards-compatible.

Pchelolo: Why? If we allow mixing inline and file-based rules the config structure would look quite ugly.. I don't really see value in backwards-compatibility except little simplification of deploying this.

d00rman: the "except" part being key. It wouldn't be possible to have it both ways. Either you define the list of files, or you use the templates stanza as before, in that order. Whichever is satisfied first, wins.

d00rman:
Hm, we will now need a way to template the rules themselves. Different URIs are used for ORES and RB, e.g., outside and inside production. We need to get them in somehow.

Pchelolo:
@d00rman I've managed to solve the problem in a quite hacky way, but it works. Not sure whether the benefits of moving most of configs out of puppet overweight the hackiness of this change. What do you think?

d00rman:
Hm hm, not bad, but indeed hacky. Let's sleep on it for now, and get the other updates out first.

Pchelolo:
After the config deploy via scap3 was implemented it's not such a pain as it's used to be, abandoning this wacky solution

PR #68, opened 2016-07-06
Upgrade to node v4.4.6

d00rman:
Bug: [T138561](https://phabricator.wikimedia.org/T138561)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 sudo: false
 
 node_js:
-  - "4.3"
+  - "4"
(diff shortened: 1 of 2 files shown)


Pchelolo:
I don't think that' required for a change like this, but 👍

d00rman:
Yes, it is. We use the node version from package.json to build the module dependencies for the deploy repo, and we have binary ones there.

Pchelolo:
@d00rman I mean if my +1 is necessary for such a change, not that the change is necessary :)

PR #70, opened 2016-07-14
Improve matching function

Pchelolo:
While rewriting change-prop to use a new driver I've noticed that sometimes our match function throws an exception since for the long list on nested properties we don't check whether all the objects along the way exist, so we get `Can't read property X of undefined` errors. 

It's not a big issue in production since we catch those errors and treat them as not-match, but it's better to fix anyway.

cc @wikimedia/services


gwicke:
:+1:

d00rman:
Nice! :+1:

PR #72, opened 2016-07-19
EPIC: Update to use new kafka driver

Pchelolo:
Updated Change-Prop to use the new kafka driver.

Major differences:
 - New driver has a `pull` model API instead of `push` as the old one. So, we have to call `consumer.consume` to get next message. This makes the `TaskQueue` obsolete - we will never receive more messages in parallel then we've called the consume method.
 - Separated retry and normal executors, inherit them from the base executor to share common functions
 - Updated tests to use new consumer/producer, improved waiting
 - The new lib is a bit slower to start up properly, so increased the test timeouts/waiting times

The config format for kafka client configuration is also significantly changed, so puppet change would be required. The changes are backwards-incompatible.

cc @wikimedia/services


gwicke:
At first glance this looks good to me. @d00rman, do you plan on doing a deeper review?

lib/base_executor.js
```diff
@@ -0,0 +1,332 @@
...
+            Math.pow(spec.retry_factor, spec.retry_limit - retryMessage.retries_left);
+        return P.delay(delay)
+        .then(() => this.hyper.post({
+            uri: '/sys/queue/events',
+            body: [ retryMessage ]
+        }));
+    }
+
+    _consumerId() {
+        return 'change-prop#' + this.rule.name;
```
d00rman: Why does this name differ from the one given in L55 ?

lib/base_executor.js
```diff
@@ -0,0 +1,332 @@
...
+            // We've got an error, but it's not from the update request, it's
+            // some bug in change-prop. Log and send a fatal error message.
+            this.hyper.log(`error/${this.rule.name}`, {
+                message: 'Internal error in change-prop',
+                error: e,
+                event: message
+            });
+            return reportError();
+        } else if (this.rule.shouldRetry(e)
+            && !this._isLimitExceeded(retryMessage, e)) {
```
d00rman: Nit: indent one more level for readability

lib/base_executor.js
```diff
@@ -0,0 +1,332 @@
...
+        })
+        .finally(this._consume.bind(this));
+    }
+
+    _notifyFinished(offset) {
+        this._pendingTaskOffsets = this._pendingTaskOffsets.filter((o) => o === offset);
+        if (this._pendingTaskOffsets.length) {
+            this._toCommit = this._pendingTaskOffsets.sort()[0];
+        } else {
+            this._toCommit = offset + 1;
```
d00rman: Why are we augmenting the offset here, but not in the block above?

lib/base_executor.js
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        if (!this._commitTimeout) {
+            this._commitTimeout = setTimeout(this._doCommit.bind(this),
+                DEFAULT_COMMIT_INTERVAL);
```
d00rman: Nit: Didn't the commit interval use to be configurable?

lib/kafka_config.js
```diff
@@ -0,0 +1,91 @@
+"use strict";
+
+const CONSUMER_DEFAULTS = {
```
d00rman: Nit: could we put some comments for these default variables and the values they hold?

lib/kafka_config.js
```diff
@@ -0,0 +1,91 @@
...
+     * @param {Object} [kafkaConf.producer] Producer configuration.
+     */
+    constructor(kafkaConf) {
+        if (!kafkaConf.metadata_broker_list) {
+            throw new Error('metadata_broker_list property is required fot the kafka config');
+        }
+
+        this._kafkaConf = kafkaConf;
+
+        const consumerTopicConf = Object.assign(CONSUMER_TOPIC_DEFAULTS,
```
d00rman: Euh, this alters the constant. The first arg to `Object#assign` is the target.

lib/kafka_config.js
```diff
@@ -0,0 +1,91 @@
...
+        return this._kafkaConf.dc_name || this._kafkaConf.produce_dc || 'datacenter1';
+    }
+
+    /**
+     * Returns a consumer configuration object
+     */
+    consumerConf(groupId, clientId) {
+        const conf = Object.assign({}, this._consumerConf);
+        conf['group.id'] = groupId;
+        conf['client.id'] = '' + (Math.random() * 1000000);
```
d00rman: This returns a float, might want to floor the result to get an int.

lib/kafka_config.js
```diff
@@ -0,0 +1,91 @@
...
+        return conf;
+    }
+
+    /**
+     * Returns a consumer configuration object
+     */
+    producerConf() {
+        const conf = Object.assign({}, this._producerConf);
+        conf['client.id'] = '' + (Math.random() * 1000000);
```
d00rman: Idem.

lib/base_executor.js
Pchelolo: Because the `_pendingTaskOffsets` holds offsets of the jobs that were not finished yet. So if it crashes now, the message under the lowest `_pendingTaskOffsets` should be reread.

If there's nothing pending, the `offset` is the one that we've just finished, so the next message to read in case of a crash is `offset + 1`

lib/base_executor.js
Pchelolo: Nope as far as I remember, and I don't see why should it be configurable?

config.example.wikimedia.yaml
```diff
@@ -30,8 +30,16 @@ spec: &spec
       x-modules:
         - path: sys/kafka.js
           options:
-            uri: 127.0.0.1:2181
+            metadata_broker_list: 127.0.0.1:9092
             dc_name: test_dc
+            produce_dc: test_dc
```
d00rman: Nit: not needed

test/feature/static_rules.js
```diff
@@ -198,35 +179,36 @@ describe('Basic rule management', function() {
...
+            "metadata.broker.list": "127.0.0.1:9092",
+            "fetch.wait.max.ms": "1",
+            "fetch.min.bytes": "1",
+            "queue.buffering.max.ms": "1"
         });
+        retryConsumer.subscribe([ 'test_dc.change-prop.retry.simple_test_rule' ]);
+        setTimeout(() => producer.produce('test_dc.simple_test_rule',
+            0,
+            JSON.stringify(common.eventWithMessageAndRandom('test', random))), 2000);
+        return retryConsumer.consume()
```
d00rman: This test is failing because this retry consumer seems to be consuming messages produced in the previous test.

Pchelolo:
Closing in favour of #81

PR #73, opened 2016-07-20
Fix a bug in error ignoring config

Pchelolo:
There was an obvious bug in our error ignoring config.

Puppet counterpart: https://gerrit.wikimedia.org/r/#/c/300166/

cc @wikimedia/services


gwicke:
:+1:

PR #75, opened 2016-07-21
Bug: ignore definition update in reaction to revision change event

Pchelolo:
When HTML content changes in RESTBase it emits 2 `resource_change` events: one for title and one for revision. The definition update rule incorrectly reacted to both. The second one treated `title/revision` string as a title and so created an additional request that resolved to 404.

cc @wikimedia/services


Pchelolo:
Puppet patch: https://gerrit.wikimedia.org/r/#/c/300442/

PR #78, opened 2016-08-10
Use one request for multiple models

Ladsgroup:
With the new version of ores live, it's much more efficient
Bug: T142360

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -254,138 +254,202 @@ spec: &spec
                         domain: ar.wikipedia.org
                       rev_by_bot: false
                     exec:
-                      uri: 'https://ores.wikimedia.org/v2/scores/arwiki/reverted/{{message.rev_id}}/?precache=true'
+                      uri: 'https://ores.wikimedia.org/v2/scores/arwiki/'
+                      query:
+                        models: 'reverted'
+                        revids: '{{message.rev_id}}'
+                        preache: true
                   - match:
                       meta:
                         domain: cs.wikipedia.org
                       rev_by_bot: false
                     exec:
-                      uri: 'https://ores.wikimedia.org/v2/scores/cswiki/reverted/{{message.rev_id}}/?precache=true'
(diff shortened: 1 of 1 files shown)


config.example.wikimedia.yaml
```diff
@@ -266,9 +266,8 @@ spec: &spec
                         domain: en.wikipedia.org
                       rev_by_bot: false
                     exec:
-                      - uri: 'https://ores.wikimedia.org/v2/scores/enwiki/damaging/{{message.rev_id}}/?precache=true'
-                      - uri: 'https://ores.wikimedia.org/v2/scores/enwiki/reverted/{{message.rev_id}}/?precache=true'
-                      - uri: 'https://ores.wikimedia.org/v2/scores/enwiki/goodfaith/{{message.rev_id}}/?precache=true'
+                      - uri: 'https://ores.wikimedia.org/v2/scores/enwiki/?models=damaging|goodfaith|reverted&
+revids={{message.rev_id}}&precache=true'
```
Pchelolo: Em, why is there a new line inside the URI? It's not supported.

In general I propose to transform all of them to the better syntax:
```yaml
exec:
    uri: 'https://ores.wikimedia.org/v2/scores/enwiki/
    query:
        models: 'damaging|goodfaith|reverted'
        revids: '{{message.rev_id}}'
        preache: true
```

And also cover the ones that were not touched by this change for consistency.

d00rman: +1 to what @Pchelolo said, especially the transformation part. It increases readability as well as correctness (encoding, etc)

Ladsgroup: The puppet change is https://gerrit.wikimedia.org/r/#/c/304125/1. Let me give it a try :)

Ladsgroup:
If this looks okay to you. I make changes to the puppet patch

Pchelolo:
Looks good to me, @Ladsgroup go ahead with puppet. 

The tests are failing because one of them is incorrect, I'll update it in a follow-up PR

PR #79, opened 2016-08-10
Fix tests and ORES typo

Pchelolo:
Follow-up for https://github.com/wikimedia/change-propagation/pull/78 - I didn't notice a typo there.

Also, after the date formatting code was improved, tests that involve date formatting need adjustment.

cc @wikimedia/services @Ladsgroup


Pchelolo:
Also, I'm skipping 'image updates' test - that feature is not enabled yet and the test is not stable.

PR #81, opened 2016-08-17
Switch to blizzard node-rdkafka driver

Pchelolo:
Switch to https://github.com/Blizzard/node-rdkafka

Some notes from labs testing:
- It works
- Perf is good
- kafka stops/restarts go unnoticeable
- if I kill the broker that's currently a leader for actively consumed partition with SIGKILL there's a little hiccup in consumption, but it comes back pretty soon

cc @wikimedia/services


lib/base_executor.js
```diff
@@ -0,0 +1,365 @@
...
+            return reportError();
+        } else if (this.rule.shouldRetry(e)
+                && !this._isLimitExceeded(retryMessage, e)) {
+            return this._retry(retryMessage);
+        } else if (!this.rule.shouldIgnoreError(e)) {
+            return reportError();
+        }
+    }
+
+    _safeParse(message) {
```
gwicke: It would be slightly more efficient if this method was synchronous. This would save one Promise allocation per successful call.

The logging case would still create a promise, but we don't need to necessarily wait for that.

lib/base_executor.js
```diff
@@ -0,0 +1,365 @@
...
+        this.consumer.consumeAsync()
+        .then((msg) => {
+            this._pendingMsgs.push(msg);
+
+            // Note: we don't return the promise here since we wanna process messages
+            // asynchronously from consuming them to be able to fill up the pendingMsg
+            // queue and achieve the level of csoncurrency we want.
+            this._safeParse(msg.message.toString('utf8'))
+            .then(this.onMessage.bind(this))
+            .finally(() => {
```
gwicke: You might want to start more consume calls even before calling onMessage. Otherwise, it looks like we'd only ever have a single operation in flight at a time.

gwicke: Also, if this.onMessage(msg) return something falsy, then you can short-cut this entirely without allocating another Promise. That creates a fast-path for high-volume applications where only few events end up matching.

lib/base_executor.js
Pchelolo: @gwicke Totally agree, but I will do this in a followup PR, wanna minimise the amount of changes in this one

lib/base_executor.js
Pchelolo: @gwicke Again, let's followup with optimisations

lib/base_executor.js
```diff
@@ -0,0 +1,365 @@
...
+                case kafka.CODES.ERRORS.ERR__PARTITION_EOF:
+                case kafka.CODES.ERRORS.ERR__TIMED_OUT:
+                    // We're reading to fast, nothing is there, slow down a little bit
+                    return P.delay(100);
+                default:
+                    this.log(`error/consumer/${this.rule.name}`, e);
+
+            }
+        })
+        .finally(() => {
```
d00rman: Why are you repeating the same block again? Shouldn't this be executed inside the `catch()` block so that it happens only on consumer errors?

lib/kafka_factory.js
```diff
@@ -91,43 +78,44 @@ class KafkaFactory {
...
+                    return reject(err);
+                }
+                // Check whether topic exists
+                consumer.getMetadata({ topic: topic }, function(err, info) {
+                    if (err) {
+                        return reject(err);
+                    }
+                    const thisTopics = info.topics.filter((t) => t.name === topic);
+                    if (!thisTopics.length || !thisTopics[0].partitions.length) {
+                        reject(new Error(`Topic ${topic} doesn't exist`));
```
d00rman: Perhaps it's time to revisit this logic and think about auto-creating the topic if it doesn't exist? Bringing down CP because a topic name was mistyped in the config looks less and less appealing to me.

package.json
```diff
@@ -39,10 +39,10 @@
     "extend": "^3.0.0",
     "hyperswitch": "^0.5.3",
     "service-runner": "^2.0.4",
-    "wmf-kafka-node": "^0.1.9",
     "json-stable-stringify": "^1.0.1",
     "htcp-purge": "^0.1.2",
-    "mediawiki-title": "^0.5.4"
+    "mediawiki-title": "^0.5.4",
```
d00rman: Need also to bump service-runner's version, as CP depends on features available only from 2.0.9

lib/base_executor.js
Pchelolo: Nono.. Basically the idea is to fill up the pending queue as fast as possible to achieve the desired level of concurrency. If you put this to the catch block, if the first message read succeed you will have concurrency 1 in the beginning since the new call to _consume will happen only after the message is processed. So if you have a large backlog is will all be processed with concurrency 1 since evert consume call will succeed until you reach the end of the partition.

lib/kafka_factory.js
Pchelolo: @d00rman the issue here is that while the old driver error out on the non-existent topic, this one (librdkafka) simply 'blocks' and does nothing, so here misconfiguring the topic name will be completely unnoticeable, so I still think fail-fast approach is better.

d00rman: But, we can read the list of topics and so know beforehand whether it exists or not. Hence, we can crete it on demand. I think this is especially of use for retry topics. That said, I also think that such a functional change is better addressed in a separate PR

Pchelolo: @d00rman by separate PR you mean leave this one intact or remove this assertion code from this one and add it in another one?

Pchelolo:
There're currently 2 issues I found so far:
 - Occasional native assertion error on startup: not a blocker, happens rarely, worker is just killed and restarted, happens only on startup. Status: investigating
 - Possibility for a MASSIVE double/triple/quadriple processing of the backlogged messages on restarts. The owner of each rule is decided by 'who has the lowest random client id', worker startup in sequence and subscribe immediately - first worker gets some messages, second worker started with lower client.id - second worker becomes an owner - second worker gets the same messages, because it all happens so fast that commit timeout didn't pass yet. Maybe we should wait for all workers to startup before subscribing to topics?

lib/kafka_factory.js
d00rman: The former :)

d00rman:
> Maybe we should wait for all workers to startup before subscribing to topics?

Yup. I was thinking about it in the general sense for `service-runner`. It may be helpful to have the master dispatch a `startup_complete` message to all workers, upon which they could then use to start subscribing to topics.

d00rman:
Hm, actually that does not really seem viable in `service-runner`. The master can easily send the message, and the worker can receive it, but there doesn't seem to exist a way for the worker to report that to the loaded services, since starting from v2, they only see the JSON-serialisable values.

Pchelolo:
> Hm, actually that does not really seem viable in service-runner. The master can easily send the message, and the worker can receive it, but there doesn't seem to exist a way for the worker to report that to the loaded services, since starting from v2, they only see the JSON-serialisable values.

Hyperswitch sees the 'server' object, so we could emit that artificial event on it. Or just add ~10 seconds delay before we subscribe. Although, 'subscribe' is synchronous and we can't start consuming messages before we subscribe, so a part of startup will become completely async. 

I'll have a look at our options..

PR #86, opened 2016-08-26
Release v0.4.1

d00rman:
Also, test the service against Node 6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 
 node_js:
   - "4"
-  - "5"
+  - "6"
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ sudo: false
 node_js:
   - "4"
   - "5"
```
Pchelolo: Let's maybe remove node 5 - we're never gonna use it in prod anyway, 4 and 6 should be ok

d00rman: Very good point, @Pchelolo

PR #87, opened 2016-08-26
Ignore 503 from ORES

Pchelolo:
ORES responds with 503 when the server is overloaded, don't make the problem worse by retrying.

@Ladsgroup Do you know any other error codes from ORES that shouldn't be retried?

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -256,6 +256,9 @@ spec: &spec
               ores_cache:
                 topic: mediawiki.revision_create
                 concurrency: 10
+                ignore:
+                  status:
+                    - 503
                 cases:
                   - match:
                       meta:
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -280,13 +280,14 @@ class BaseExecutor {
                 event: message
             });
             return reportError();
-        } else if (this.rule.shouldRetry(e)
-                && !this._isLimitExceeded(retryMessage, e)) {
-            return this.hyper.post({
-                uri: '/sys/queue/events',
-                body: [ retryMessage ]
-            });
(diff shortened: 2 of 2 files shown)


Ladsgroup:
I think I know why it's overloading and that will be fixed in Monday. If a kind of throttling can be applied,  like when it errors, wait for n seconds and then retry, it would be amazing. If it's not possible, I guess just ignoring won't do much harm (specially since overload percentage is less than 1%)

Pchelolo:
@Ladsgroup There's throttling already, it retries after 1 minute and if it fails again - retries after 6 minutes. It's not fixing any issue we see currently, just I thought that in case ORES would be experiencing issues and the rate of 503 would become significant, retrying them (even after 1 minute) only worsened the problem. Basically I wanna protect from potential situation where ORES is 'almost dead' sending lots of 503 and CP kills it completely with retries. Since the cost of missing an update is minimal I think it's better to ignore 503 responses.

Currently it will retry 5xx except 503.

Ladsgroup:
Here's the monitoring graph for overload error, https://grafana.wikimedia.org/dashboard/db/ores?panelId=9&fullscreen

They are not much even with a big pressure we are currently putting on it, so I really have no preferences.

Pchelolo:
@Ladsgroup ok. I think I'll still push this in for extra safety

Ladsgroup:
:+1:

gwicke:
@Pchelolo, this might be an example where processing main requests & retries in the same worker & sharing the concurrency limit between them might be a win. This would eliminate the risk of doubling concurrency relative to normal processing during times of high retry rates.

Pchelolo:
Corresponding gerrit: https://gerrit.wikimedia.org/r/307810

PR #88, opened 2016-08-30
Wikidata: Correctly boolean-ise the `success` flag

d00rman:
The `success` flag in the result of `wbgetentities` is set to `1` if the
query has been successfully executed, so double-negate when checking
whether we should extract the results or not.


d00rman:
cc @wikimedia/services

PR #89, opened 2016-08-30
Wikidata: Make sure we have the response we need to process the deps

d00rman:
While looking at the returned `success` flag tells us whether the query
has been successfully executed, it does not ensure that the response
contains the fields we need to send resource_change events for the
dependencies, so check for them explicitly in the response.

### DIFF
--- a/sys/dep_updates.js
+++ b/sys/dep_updates.js
@@ -88,7 +88,20 @@ function createWikidataTemplate(options) {
             }
         })),
         resourceChangeTag: 'wikidata',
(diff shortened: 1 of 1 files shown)


d00rman:
cc @wikimedia/services

d00rman:
The lack of this check is causing `Internal error in Change Prop` errors ( e.g. https://www.wikidata.org/w/api.php?format=json&formatversion=2&action=wbgetentities&ids=Q17343313&props=sitelinks%2furls&normalize=true ), so self-merging and deploying. We can refine it at a later point.

PR #90, opened 2016-08-31
Wikidata: Only log Wikibase errors and ignore non-mainspace titles

d00rman:
Two improvements for mild issues seen in production:
- do not log all of the events that do not produce links to WP articles, as not all items are actually linked on a site, just log errors returned by Wikibase
- only wikidata items can have site links, so disregard any non-main namespace titles


d00rman:
cc @wikimedia/services

PR #93, opened 2016-09-01
Switch to new even formats and topics

Pchelolo:
Together with analytics we've changed the event schemas and started producing events to the new topics, so updated CP to use these new updated events so that eventually we could stop producing deprecated events.

cc @wikimedia/services


Pchelolo:
Corresponding puppet change: https://gerrit.wikimedia.org/r/#/c/308077/

d00rman:
:+1: I'm guessing we have to wait till the end of next week to deploy this?

d00rman:
Let's put this on hold until late next week.

Pchelolo:
> 👍 I'm guessing we have to wait till the end of next week to deploy this?
Why? New events are in prod right now.

d00rman:
> Why? New events are in prod right now.

Let's focus on tranclusions now. If all goes well, we can get this change in a PuppetSWAT window.

PR #94, opened 2016-09-02
Rerender mobile-sections when wikidata item is changed

Pchelolo:
https://phabricator.wikimedia.org/T144632

cc @wikimedia/services


gwicke:
LGTM, but leaving merge to @d00rman

Pchelolo:
Here's corresponding gerrit: https://gerrit.wikimedia.org/r/#/c/308230/

PR #97, opened 2016-09-06
Added if-unmodified-since to transcludes

Pchelolo:
cc @wikimedia/services


d00rman:
Puppet change is [Gerrit 308773](https://gerrit.wikimedia.org/r/#/c/308773/)

config.example.wikimedia.yaml
```diff
@@ -101,6 +101,8 @@ spec: &spec
                 match:
                   meta:
                     uri: '/^(?<proto>https?):\/\/[^\/]+\/api\/rest_v1\/page\/html\/(?<title>[^/]+)$/'
+                  tags:
+                    - restbase
```
Pchelolo: @d00rman Since you've made the puppet change, could you add this to your change too?

PR #98, opened 2016-09-07
Deduplicate continuation events

Pchelolo:
We can deduplicate some of the continuation events when some templates are edited quickly concequtively.

Bug: https://phabricator.wikimedia.org/T145011
cc @wikimedia/services

### DIFF
--- a/sys/dep_updates.js
+++ b/sys/dep_updates.js
@@ -8,6 +8,7 @@ const Title = require('mediawiki-title').Title;
 
 const BACKLINKS_CONTINUE_TOPIC_NAME = 'change-prop.backlinks.continue';
 const TRANSCLUDES_CONTINUE_TOPIC_NAME = 'change-prop.transcludes.continue';
+const DEDUPE_LOG_SIZE = 100;
 
 function createBackLinksTemplate(options) {
     return {
@@ -134,6 +135,7 @@ class DependencyProcessor {
                 siprop: 'general|namespaces|namespacealiases'
             }
         }));
+        this.latestMessages = [];
     }
 
     setup(hyper) {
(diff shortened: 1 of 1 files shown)


sys/dep_updates.js
```diff
@@ -177,13 +179,56 @@ class DependencyProcessor {
...
+    _isDuplicate(message) {
+        if (!message.continue) {
+            // Not a continuation event, don't deduplicate
+            return false;
+        }
+
+        const currEventInfo = {
+            domain: message.meta.domain,
+            title: message.original_event.page_title,
+            id: message.original_event.meta.id,
```
d00rman: Did you mean(TM) `meta.request_id` ?

Pchelolo: No, I meant `message.original_event.meta.id`, but for our current purpose it doesn't matter whether to compare id or request_id.

sys/dep_updates.js
```diff
@@ -177,13 +179,56 @@ class DependencyProcessor {
...
+        }
+
+        if (this.latestMessages.some((oldEvent) => {
+                return oldEvent.domain === currEventInfo.domain
+                    && oldEvent.title === currEventInfo.title
+                        // This represents forking - we've consumed some continuation twice,
+                        // so we've got exact same event processed recently
+                    && (oldEvent.id === currEventInfo.id && oldEvent.sequence_num === currEventInfo.sequence_num
+                        // This represents deduplication on rapid consequent template changes
+                        || oldEvent.id !== currEventInfo.id && oldEvent.sequence_num <= currEventInfo.sequence_num);
```
d00rman: Here we are comparing an affected page from two different events. So why are their sequence numbers relevant? It kind of depends on the ordering and the pace each of these has been consumed. If we have already re-rendered a page, I am not sure we can positively state that we don't need to re-render it again. Or am I missing something?

Pchelolo: @d00rman The sequence number matters precisely for the reason you've voiced. We only wanna drop the older continuation batch, not the newer one. Here's an example:
1. Template edit happens, an event is posted with ID=1
2. Several batches processed, so now we're on ID=1,SEQ=5
3. Another template edit happens, and event posted with ID=2
4. A couple of batches processed started by ID=2 edit, so we've got ID=2,SEQ=1 in the log
5. An old batch is still progressing, so we've received a message with ID=1,SEQ=6. We look at the recent events log and see, that there's been a newer batch job started - that one will certainly get to our position soon, so we can drop current (old) job - we drop it.

d00rman: Doesn't this assume that when we sent the first re-render request, the second template edit had already happened? This is the part I'm not clear about.

Pchelolo: @d00rman No, it doesn't. It will deduplicate eventually after the second template edit happens and the new batch of continue events will start to kick off. It might still process a couple of continue events from the old batch (corresponding to the first edit), but we don't really care. What I really don't want to happen is false-positives (dedupe something we shouldn't) and in this scheme it shouldn't happen.

PR #101, opened 2016-09-09
Don't count rejected messages against the concurrency limits

Pchelolo:
We only need to count actually processed messages agains the concurrency limit, if the message is rejected - don't push it to the pending queue.

Since we have lot's of rejections this might make us go a bit faster when we need.

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -66,13 +66,17 @@ class BaseExecutor {
         this._consuming = true;
         this.consumer.consumeAsync()
         .then((msg) => {
-            this._pendingMsgs.push(msg);
+            // We're pushing it to pending messages only if it matched so that items
+            // that don't match don't count against the concurrency limit.
 
             // Note: we don't return the promise here since we wanna process messages
             // asynchronously from consuming them to be able to fill up the pendingMsg
             // queue and achieve the level of csoncurrency we want.
             this._safeParse(msg.message.toString('utf8'))
-            .then(this.onMessage.bind(this))
+            .then((message) => this.onMessage(message, () => {
+                // This is called when message passe all tests and about to begin execution
+                this._pendingMsgs.push(msg);
+            }))
             .finally(() => {
                 this._notifyFinished(msg);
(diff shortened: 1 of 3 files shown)


lib/base_executor.js
```diff
@@ -66,13 +66,17 @@ class BaseExecutor {
...
+            // We're pushing it to pending messages only if it matched so that items
+            // that don't match don't count against the concurrency limit.
 
             // Note: we don't return the promise here since we wanna process messages
             // asynchronously from consuming them to be able to fill up the pendingMsg
             // queue and achieve the level of csoncurrency we want.
             this._safeParse(msg.message.toString('utf8'))
-            .then(this.onMessage.bind(this))
+            .then((message) => this.onMessage(message, () => {
+                // This is called when message passe all tests and about to begin execution
```
gwicke: Minor: passes

lib/base_executor.js
```diff
@@ -66,13 +66,17 @@ class BaseExecutor {
...
-            this._pendingMsgs.push(msg);
+            // We're pushing it to pending messages only if it matched so that items
+            // that don't match don't count against the concurrency limit.
 
             // Note: we don't return the promise here since we wanna process messages
             // asynchronously from consuming them to be able to fill up the pendingMsg
             // queue and achieve the level of csoncurrency we want.
             this._safeParse(msg.message.toString('utf8'))
-            .then(this.onMessage.bind(this))
+            .then((message) => this.onMessage(message, () => {
```
gwicke: This callback mechanism strikes me as a bit convoluted. Could we make the matching synchronous instead? That would also speed up matching by avoiding to allocate promises for skipped messages.

Pchelolo: Ye, I agree not the cleanest solution, just wanted to do it with minimal code changes for now to deploy and test whether it actually makes difference. But I can revisit this and make improve it right away

Pchelolo: @gwicke so, data shows that this indeed helped, I'm going to refactor it a little bit to get rid of an ugly callback and add a 'dequeue speed' metrics

PR #103, opened 2016-09-09
Don't create a promise on JSON.parse, avoid ugly onProcess callback

Pchelolo:
Follow-up on https://github.com/wikimedia/change-propagation/pull/101

- Some generic refactoring
- More sync processing until we verify that the message will indeed be processed - less promises, should save some perf/mem
- Added a stat on dequeue speed

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -66,23 +66,34 @@ class BaseExecutor {
         this._consuming = true;
         this.consumer.consumeAsync()
         .then((msg) => {
-            // We're pushing it to pending messages only if it matched so that items
-            // that don't match don't count against the concurrency limit.
+            const payload = msg.message.toString('utf8');
+            let message;
+            try {
+                message = JSON.parse(payload);
+            } catch (e) {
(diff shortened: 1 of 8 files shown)


lib/retry_executor.js
```diff
@@ -37,38 +41,28 @@ class RetryExecutor extends BaseExecutor {
         }
     }
 
-    onMessage(message, onExec) {
+    getOptionIndex(message) {
```
gwicke: How about something like `shouldProcessMessage`, or `handlerIndex`?

lib/base_executor.js
```diff
@@ -270,6 +284,7 @@ class BaseExecutor {
     }
 
     _catch(message, retryMessage, e) {
+        console.log(e);
```
d00rman: Ups :)

d00rman:
Very nice! One minor nit in-lined. How about bumping the version too since it's a performance improvement?

d00rman:
I'll bump the version and get rid of the `console.log()` in a subsequent PR

PR #105, opened 2016-09-12
Use separate topic for transclusions

Pchelolo:
Depends-On: https://gerrit.wikimedia.org/r/310042
Bug: https://phabricator.wikimedia.org/T145040
cc @wikimedia/services


d00rman:
Since now we have a dedicated topic, we should get rid of [the `change-prop` tag](https://github.com/wikimedia/change-propagation/blob/fa240648f1c6c807546bef991a33f2e9217bcf97/sys/dep_updates.js#L330), in this way speeding up the execution to some extent.

Pchelolo:
@d00rman Done

PR #106, opened 2016-09-14
Fix: The correct topic name is revision-visibility-change

d00rman:


d00rman:
cc @wikimedia/services

d00rman:
[Fix for prod](https://gerrit.wikimedia.org/r/#/c/310555/) already deployed.

PR #107, opened 2016-09-19
Create separate processors for transcludes-driven rerenders

Pchelolo:
We could potentially do something programmatically to avoid config copy-pasta, but I think it doesn't worth it, being explicit is cleaner in this case.

Bug: https://phabricator.wikimedia.org/T145804
cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -43,60 +43,49 @@ spec: &spec
             concurrency: 250
             templates:
 
-              summary_rerender: &summary_rerender_spec
+              summary_definition_rerender: &summary_definition_rerender_spec
                 topic: resource_change
                 retry_limit: 2
                 retry_delay: 500
                 retry_on:
                   status:
                     - '5xx'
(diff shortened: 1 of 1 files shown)


Pchelolo:
Corresponding change in deploy repo: https://gerrit.wikimedia.org/r/311602

config.example.wikimedia.yaml
```diff
@@ -43,7 +43,7 @@ spec: &spec
             concurrency: 250
             templates:
 
-              summary_rerender:
+              summary_rerender: &summary_rerender_spec
```
d00rman: Let's take this opportunity to consolidate `summary_rerender` and `definition_rerender` using the `cases` stanza. As a bonus, we get out of it 0 new rules for them.

Pchelolo:
@d00rman done

config.example.wikimedia.yaml
Pchelolo: done

PR #108, opened 2016-09-19
Give ORES it's own module

Pchelolo:
There are talks about adding a topic with 'bad edits' Do do this we need to
1. Update ORES
2. If ORES rates are bad - construct and emit a new event to the bad edits topic.

It's not quite possible to do this purely in the config - we need conditional, complex request-handler support etc. I don't think it worths adding all those capabilities, so give ORES it's own JS module.

cc @wikimedia/services

### DIFF
--- a/sys/ores.js
+++ b/sys/ores.js
@@ -16,7 +16,7 @@ class OresUpdater {
         const rev_id = req.body.rev_id;
         if (!this.options.models[wiki]) {
             throw new HTTPError({
-                status: 200,
+                status: 400,
                 body: {
                     message: `ORES precache request for unmodified wiki ${wiki}`
                 }
--- a/sys/ores.js
+++ b/sys/ores.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const utils = require('../lib/utils');
+const HTTPError = require('hyperswitch').HTTPError;
 
 class OresUpdater {
     constructor(options) {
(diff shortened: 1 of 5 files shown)


sys/ores.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    updateORES(hyper, req) {
+        const wiki = req.body.database;
+        const rev_id = req.body.rev_id;
+        if (!this.options.models[wiki]) {
+            hyper.log('warn/ores', {
+                message: `ORES precache request for unmodified wiki ${wiki}`
+            });
+            return { status: 200 };
```
d00rman: The point of converting the ORES rules into a module is to be able to retrieve its response, but in this instance we do not actually have one, so we should probably use a different code in this case. `400` seems a good one.

stephanebisson:
Does this make the ORES response available? How would it be used is subsequent `exec` blocks?

Pchelolo:
@stephanebisson Yes. Now we will code up the subsequent requests in JavaScript.

Pchelolo:
This patch is waiting for all the rest of the EventBus infrastructure to settle in first:
 - https://gerrit.wikimedia.org/r/#/c/312277/
 - https://gerrit.wikimedia.org/r/#/c/312274/
 - Defining a schema for the review stream topic
 - creating the review stream topic

Then this will be updated to actually emit something in that topic.

Pchelolo:
Not needed anymore, we've completely changed how ORES is precached.

PR #111, opened 2016-09-21
Set x-restbase-mode header

Pchelolo:
RestbaseUpdateJobs used to set the `x-restbase-mode` header to indicate a template/file reparse that was sent to parsoid and it applied some optimisations. Apparently it's not used in parsoid, but let's feature-match RBUpdateJobs as parsoid will eventually implement this.

We need to set it in CP too

CC @wikimedia/services


d00rman:
Nice find! :100:

d00rman:
Config change for the deploy repo is [Gerrit 312125](https://gerrit.wikimedia.org/r/#/c/312125/).

PR #112, opened 2016-09-29
First iteration on the private-wiki support

Pchelolo:
It works, but lacks some features still, mostly related to error handling and handling of situations when our session expired. Will add them in the next iteration.

We'd need to set up a superuser for each of those private wikis we wanna support.

Bug: https://phabricator.wikimedia.org/T137140
cc @wikimedia/services


config.example.wikimedia.yaml
```diff
@@ -1,12 +1,25 @@
 spec: &spec
   x-sub-request-filters:
+    - path: lib/auth_http_filter.js
+      options:
+        username: 'SOME_USER_WILL_BE_HERE'
+        password: 'SOME_PASSWORD_WILL_BE_HERE'
```
d00rman: Note to ourselves: before merging and deploying this change, we will need to set up the account information in `ops/private` so as not to publicly advertise the real credentials.

Pchelolo:
I wanna put this one on hold. I've discovered an alternative way to authenticate a bot on wiki which might be a better way forward here. See https://phabricator.wikimedia.org/T137140#2700822

PR #113, opened 2016-09-29
Go back to blizzard driver since bugs that blocked us weer resolved

Pchelolo:
We've used our own fork because of this bug: https://github.com/Blizzard/node-rdkafka/issues/19

The `rebalance_cb` is not set in upstream by default any more, so we can safely switch back to upstream version. I've tested it for a while and verified it's all right.

cc @wikimedia/services


d00rman:
LGTM, but let's merge and deploy next week.

PR #116, opened 2016-10-05
Release v0.5.3

d00rman:
Also update dependencies and signal that this service works only on Node 4 and above.


d00rman:
cc @wikimedia/services

Pchelolo:
+1

d00rman:
Published as [v0.5.3](https://github.com/wikimedia/change-propagation/tree/v0.5.3)

PR #118, opened 2016-10-06
React to page_properties event

Pchelolo:
Once we switch to this event we'd need to start reacting to it instead of the current hack.


d00rman:
Topics created, good to go. We have to remember to remove the resource_change rule once [Gerrit 314605](https://gerrit.wikimedia.org/r/#/c/314605/) is in prod.

PR #119, opened 2016-10-13
Check that value we are testing with a regex is actually a string

Pchelolo:
JavaScript keeps surprising, `/.+/.test(undefined) === true`.  So we need to check that the value we're testing with a regex is actually a string.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -592,25 +592,21 @@ spec: &spec
                 # We don't support 'OR' in the match section, so workaround it by 2 cases with identical exec
                 cases:
                   - match:
-                      meta:
-                        uri: '/^https?:\/\/[^\/]+\/wiki\/(?<title>.+)$/'
                       added_properties:
                         page_image: '/.+/' # Regex that matches anything just to check the prop is set
                     exec:
                       method: get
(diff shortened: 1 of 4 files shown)


config.example.wikimedia.yaml
```diff
@@ -592,25 +592,21 @@ spec: &spec
...
                         redirect: false
                   - match:
-                      meta:
-                        uri: '/^https?:\/\/[^\/]+\/wiki\/(?<title>.+)$/'
                       removed_properties:
                         page_image: '/.+/' # Regex that matches anything just to check the prop is set
                     exec:
                       method: get
-                      uri: 'https://{{message.meta.domain}}/api/rest_v1/page/summary/{{match.meta.uri.title}}'
+                      uri: 'https://{{message.meta.domain}}/api/rest_v1/page/summary/{message.page_title}'
```
gwicke: How is `message.page_title` encoded? If it is underscored, but might contain bare slashes, then I think we should use `{{message.page_title}}`.

Pchelolo: @gwicke message.page_title is not uriEncoded, so we use single curly braces to uriEncode it here.

gwicke: @Pchelolo, good point, I stand corrected. My template-fu is evidently rusty.

PR #120, opened 2016-10-13
Narrow down wikidata description updates

Pchelolo:
Almost 2/3 of the load on MCS cluster is due to wikidata-description updates. Before, to be on the safe side, we've been rerendering all summaries/mobile content whenever a wikidata item's changed. But actual description changes are like 1% of them, so we doing lots of extra work. 

Good news is there's a way to rule out most of the unnesessary updates by looking at the edit comment - it's impossible to modify that comment when edit is made, so they're pretty stable.
It's a hack solution, because those comments are not a part of the official wikidata API or contract, so I'm proposing better, more general solutions in the ticket, but that might take time, and we need to lower the load right now without waiting for that.

Another thing is that if a wikidata item revision is suppressed, we might need to rerender summary/mobile as well, because it's quite likely that some description was vandalised and the revision suppression is a result of that vandalism. The rate of revision suppressions is super low, so we can avoid doing any more fine-grained filtering here.

Bug: https://phabricator.wikimedia.org/T148079

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -539,8 +539,22 @@ spec: &spec
                   headers:
                     cache-control: no-cache
 
-              wikidata_description:
+              wikidata_description_on_edit:
                 topic: mediawiki.revision-create
+                match:
+                  meta:
+                    domain: www.wikidata.org
+                  page_namespace: 0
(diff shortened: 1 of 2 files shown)


Pchelolo:
Deploy repo change: https://gerrit.wikimedia.org/r/#/c/315866/

PR #123, opened 2016-10-19
Rerender mobile-sections when page images are changed

Pchelolo:
The [mobile-sections-lead endpoint](https://en.wikipedia.org/api/rest_v1/page/mobile-sections-lead/Tank) we expose a page image, same way as in the summary, so the same race condition is possible as for the summary.

cc @wikimedia/services @wikimedia/mobile


Pchelolo:
Deploy repo gerrit: https://gerrit.wikimedia.org/r/316912

PR #125, opened 2016-10-21
Added description to the commit error

Pchelolo:
The `Error` object created by `node-rdkafka` is not being rendered correctly by logger, so call `toString` to add description explicitly.

Bug: https://phabricator.wikimedia.org/T148820


Pchelolo:
I don't think that adding a `stack` property would be useful here, because the errors are generated in native code in kafka bindings, so wouldn't have much value for us.

PR #126, opened 2016-10-21
Improved logging levels

Pchelolo:
Our logs for change-prop are full of `retry count exceeded` messages, which are quite useless. If we wanna see the events for which the retry count was exceeded, we can go to kafka error topic.

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -275,7 +275,7 @@ class BaseExecutor {
      */
     _isLimitExceeded(message, e) {
         if (message.retries_left <= 0) {
-            this.log(`error/${this.rule.name}`, {
+            this.log(`warn/${this.rule.name}`, {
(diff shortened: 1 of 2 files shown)


lib/base_executor.js
```diff
@@ -275,7 +275,7 @@ class BaseExecutor {
      */
     _isLimitExceeded(message, e) {
         if (message.retries_left <= 0) {
-            this.log(`error/${this.rule.name}`, {
+            this.log(`warn/${this.rule.name}`, {
```
d00rman: These will still get logged - we use `info` locally and `warn` for logstash

d00rman:
Merging and deploying to include the `siteInfo` addition, but ideally we'd either need to decrease the logging level to `debug` and just use sampled logging for this.

PR #127, opened 2016-10-26
Ban wikidata updates

Pchelolo:
We wanna remove RESTBase from wikidata, so ban wikidata RESTBase updates. I'll create a corresponding deploy repo patch as we settle on this.

Bug: https://phabricator.wikimedia.org/T149114
cc @wikimedia/services


Pchelolo:
I think there's a general consensus on the task to remove RESTBase from wikidata, so we can move on this now.

PR #128, opened 2016-10-28
Added different comment regex for the wikidata description change

Pchelolo:
Bug: https://phabricator.wikimedia.org/T149456
@wikimedia/services @wikimedia/mobile

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -547,7 +547,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/(?:wbeditentity)|(?:wbsetdescription)/'
+                  comment: '/wbeditentity|wbsetdescription/'
                 exec:
                   method: 'post'
                   uri: '/sys/links/wikidata_descriptions'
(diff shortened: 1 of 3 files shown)


Pchelolo:
Deploy repo patch:  https://gerrit.wikimedia.org/r/318657

config.example.wikimedia.yaml
```diff
@@ -547,7 +547,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/wbeditentity/'
+                  comment: '/(?:wbeditentity)|(?:wbsetdescription)/'
```
d00rman: Why not simply: `/wbeditentity|wbsetdescription/` ?

sys/dep_updates.js
```diff
@@ -107,7 +107,7 @@ function createWikidataTemplate(options) {
             const siteLinks = res.body.entities[Object.keys(res.body.entities)[0]].sitelinks;
             return Object.keys(siteLinks).map((siteId) => {
                 const pageURI = siteLinks[siteId].url;
-                const match = /^https?:\/\/([^\/]+)\/wiki\/(.+)$/.exec(pageURI);
+                const match = /^https?:\/\/([^/]+)\/wiki\/(.+)$/.exec(pageURI);
```
gwicke: Interesting that the slash doesn't need escaping when used inside of a char class. FF parses it the same way though, so I guess it's not just a syntax quirk in v8.

gwicke:
LGTM. @Pchelolo, any reason *not* to merge this now?

Pchelolo:
@gwicke There is a discussion https://phabricator.wikimedia.org/T149456#2753477 on whether it's possible to avoid the hack and implement a proper solution, but I think it's better to deploy this now to unblock the Android app team to unblock them in their effort on wikidata description editing in the app.

PR #130, opened 2016-11-03
Ignore updates from all Wikidata domains

d00rman:
In production there are www.wikidata.org and test.wikidata.org. We want
to ignore both of them when it comes to updates.

Bug: [T149114](https://phabricator.wikimedia.org/T149114)


d00rman:
cc @wikimedia/services

d00rman:
Self-merging to deploy right now.

PR #131, opened 2016-11-03
Only restrict wikidata to certain namespaces

Pchelolo:
We need to allow visualeditor on all namespaces except 0 and 120

cc @wikimedia/services


d00rman:
Why leaving all purges for the whole domain?

Pchelolo:
@d00rman Purges are based on the `resource-change` event where we have no `page_namespace` property.

PR #132, opened 2016-11-03
Change the endpoint we call for page restore

Pchelolo:
After we switch to restriction-table based access checking we'd need to call a different endpoint to signal an undelete. The revision-table based wouldn't be broken too, so it's safe to deploy prior to the switch.

The problem is that on the HTML endpoint the access check is done in parallel with content request, and the content request is the one that would update the restriction table, so we get a race condition. Notifying the `title` endpoint avoids that issue.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -261,7 +261,7 @@ spec: &spec
                     domain: /\.wikidata\.org$/
                 exec:
                   method: get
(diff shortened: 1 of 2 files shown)


Pchelolo:
Created a gerrit patch: https://gerrit.wikimedia.org/r/#/c/320458/

PR #133, opened 2016-11-07
Made logging more consistent

Pchelolo:
We need to emit consistent types of properties at least within the service. Also, it doesn't make sense to emit the original even as object, we never index it anyway. Stringified version is OK.

Bug: https://phabricator.wikimedia.org/T150195

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -150,7 +150,7 @@ class BaseExecutor {
                         this.log(`error/commit/${this.rule.name}`, {
                             msg: 'Commit failed',
                             offset: committing.offset,
-                            raw_event: JSON.stringify(committing.message),
+                            raw_event: utils.stringify(committing.message),
                             description: e.toString()
                         });
                     });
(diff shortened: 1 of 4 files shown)


lib/base_executor.js
```diff
@@ -150,7 +150,7 @@ class BaseExecutor {
                         this.log(`error/commit/${this.rule.name}`, {
                             msg: 'Commit failed',
                             offset: committing.offset,
-                            event: committing.message.toString(),
+                            raw_event: JSON.stringify(committing.message),
```
d00rman: Let's surround the stringifying with a `try/catch` block to avoid any possible problems. Idem for the other places.

d00rman: Also, why have you changed here to `raw_event` but all the others are kept as `event` ? I don't think ES will like that.

Pchelolo:
@d00rman Updated.

d00rman:
Travis is slow, giving up and merging.

PR #134, opened 2016-11-15
Upgraded to node-rdkafka version 0.6.0

Pchelolo:
The new `node-rdkafka` version came out, let's upgrade. It has some fixes we've needed and changes API a bit, that's why the patch is so huge. 

Also, this one is still on `librdkafka@0.9.1`, so no need to update the apt package for this particular PR.

Tested with unit tests and gave it some artificial load locally with kafkacat, all looks good. 

A major change is that `produce` is sync now, but it's only waiting for the message to get to the internal `librdkafka` queue. We might consider using the delivery callback and mangling the `require_acks` config parameter to make the delivery guaranteed, but that' a separate task, right now it works exactly as it used to.

Another major change is that now it doesn't hang on `consumer.disconnect()` call, so we can finally implement graceful shutdown for the service. It's a pretty big deal, because on `disconnect` consumer flushes it's internal commit queue and producer flushes the produce queue, so we don't loose commits or messages.

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -66,7 +66,7 @@ class BaseExecutor {
         this._consuming = true;
         this.consumer.consumeAsync()
         .then((msg) => {
-            const payload = msg.message.toString('utf8');
+            const payload = msg.value.toString('utf8');
             let message;
 
             this.hyper.metrics.increment(`${this.statName}_dequeue`, 1, 0.1);
@@ -150,7 +150,7 @@ class BaseExecutor {
                         this.log(`error/commit/${this.rule.name}`, {
                             msg: 'Commit failed',
                             offset: committing.offset,
-                            raw_event: utils.stringify(committing.message),
+                            raw_event: utils.stringify(committing.value),
                             description: e.toString()
                         });
                     });
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -354,6 +354,10 @@ class BaseExecutor {
         return errorEvent;
     }
 
+    close() {
+        this.consumer.disconnect();
+    }
+
     static decodeError(e) {
         if (Buffer.isBuffer(e.body)) {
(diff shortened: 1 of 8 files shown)


Pchelolo:
Something is wrong with travis change-prop tests. They finish but the job doesn't... I've tried to run travis tests on current master, but the same issue happens. So, let's wait and see if travis fixes itself.

Also, this now depends on https://github.com/wikimedia/htcp-purge/pull/6

Pchelolo:
Something is wrong with travis - the build finishes but travis never realises it. It works locally nicely, I've even downloaded travis docker image and tried to reproduce all steps locally - all looks good. 

At first it seemed like killing kafka/zookeeper after running tests helped, but nope, sometimes the build is still stalled. I'm starting to think it's a travis issue, let's wait a bit to see if it resolves itself.

lib/base_executor.js
```diff
@@ -141,7 +141,7 @@ class BaseExecutor {
                     const committing = this._toCommit;
                     return this.consumer.commitAsync(committing)
                     .then(() => {
-                        if (this._toCommit.offset === committing.offset) {
+                        if (this._toCommit && this._toCommit.offset === committing.offset) {
```
Pchelolo: It's done async, so by the time this's executed `_toCommit` might already have been nullified.

lib/base_executor.js
```diff
@@ -150,7 +150,7 @@ class BaseExecutor {
                         this.log(`error/commit/${this.rule.name}`, {
                             msg: 'Commit failed',
                             offset: committing.offset,
-                            raw_event: utils.stringify(committing.message),
+                            raw_event: committing.value.toString(),
```
Pchelolo: The value here is a Buffer, so JSON.stringify doesn't work for it.

gwicke:
LGTM.

PR #136, opened 2016-11-16
Use delivery callback to guarantee delivery

Pchelolo:
After we've updated to `node-rdkafka@0.6.0` we finally can use the delivery report to track message production up to kafka broker. Now production becomes guaranteed for real, not just until entering the librdkafka internal production buffer.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -27,13 +27,22 @@ class GuaranteedProducer extends kafka.Producer {
     /**
      * @inheritDoc
      */
-    constructor(conf, topicConf) {
+    constructor(conf, topicConf, logger) {
         super(conf, topicConf);
+        this._logger = logger;
         this._pending = {};
 
         this.on('delivery-report', (report) => {
             const reportKey = `${report.topic}:${report.key}`;
             const resolve = this._pending[reportKey];
+            if (!resolve) {
+                this._logger.log('error/produce', {
+                    message: 'Unknown resolver in delivery report',
+                    report
+                });
+                return;
+            }
+            
             delete this._pending[reportKey];
             resolve(report);
         });
@@ -164,11 +173,13 @@ class KafkaFactory {
         });
     }
 
-    _createProducerOfClass(ProducerClass) {
+    _createProducerOfClass(ProducerClass, logger) {
         return new P((resolve, reject) => {
             const producer = new ProducerClass(
                 this._producerConf,
-                this._producerTopicConf);
+                this._producerTopicConf,
+                logger
+            );
             producer.once('error', reject);
             producer.connect(undefined, (err) => {
                 if (err) {
@@ -180,12 +191,12 @@ class KafkaFactory {
 
     }
 
-    createProducer() {
-        return this._createProducerOfClass(kafka.Producer);
+    createProducer(logger) {
+        return this._createProducerOfClass(kafka.Producer, logger);
     }
 
-    createGuaranteedProducer() {
-        return this._createProducerOfClass(GuaranteedProducer);
+    createGuaranteedProducer(logger) {
+        return this._createProducerOfClass(GuaranteedProducer, logger);
     }
 }
 module.exports = KafkaFactory;
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -42,7 +42,7 @@ class GuaranteedProducer extends kafka.Producer {
                 });
                 return;
             }
-            
+
             delete this._pending[reportKey];
             resolve(report);
         });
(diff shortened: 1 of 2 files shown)


lib/kafka_factory.js
```diff
@@ -20,9 +20,74 @@ const PRODUCER_DEFAULTS = {
...
+    /**
+     * @inheritDoc
+     */
+    constructor(conf, topicConf) {
+        super(conf, topicConf);
+        this._pending = {};
+
+        this.on('delivery-report', (report) => {
+            const reportKey = `${report.topic}:${report.key}`;
+            const resolve = this._pending[reportKey];
```
gwicke: Are we guaranteed to only ever get a single report? If not, then it would make sense to check whether we still have an outstanding request for the key before calling resolve.

lib/kafka_factory.js
```diff
@@ -20,9 +20,74 @@ const PRODUCER_DEFAULTS = {
...
+            delete this._pending[reportKey];
+            resolve(report);
+        });
+
+        this._pollInterval = setInterval(() => this.poll(), 500);
```
gwicke: At some point it might be nice to poll only when there are requests outstanding, and perhaps use shorter intervals when that is the case.

lib/kafka_factory.js
Pchelolo: @gwicke Yea, it's guaranteed. There's a similar structure in the C++ driver code that stored callbacks mapped by the key, so it's impossible to get multiple delivery-report events. At least if I'm readying the code correctly.

gwicke: Do you think it would be worth adding a sanity check in any case? Logging such a failure would probably be nicer than blowing up the worker.

Pchelolo: Ye, you're right. Added logging

lib/kafka_factory.js
gwicke: Yup, the current approach is definitely simpler. That's why I said "at some point" ;)

lib/kafka_factory.js
```diff
@@ -99,15 +173,30 @@ class KafkaFactory {
...
+                if (err) {
+                    return reject(err);
+                }
+                return resolve(producer);
+            });
         });
+
+    }
+
+    createProducer(logger) {
```
gwicke: Do we have a use case for best-effort production?

gwicke: @Pchelolo, do we actually *need* unreliable production for testing, or is it just that we don't necessarily need reliable production there?

Pchelolo: @gwicke The latter. We just don't need reliability, and non-reliable production is faster, so the tests are faster

gwicke:
Looks nice & clean overall. Nice work!

lib/kafka_factory.js
gwicke: @Pchelolo, is the slow-down necessarily that large, or is it mostly caused by waiting for the 500ms timer to fire? If it's the latter, then maybe that could be a hint for tweaking the interval.

lib/kafka_factory.js
Pchelolo: @gwicke Lemme experiment.

Pchelolo: On my laptop even with 50 ms polling it's slowing down the tests quite significantly.

gwicke: Just out of curiosity, what are the absolute test run times with 500ms / 50ms?

Pchelolo: @gwicke testing this requires quite some changes in the tests and I've already git checkouted these temp stuff.. Sorry. If you are *really* curious I can code it up again..

gwicke: Don't worry, just wondered about the order of magnitude we are talking about here.

Pchelolo: @gwicke Basically in prod the  number wouldn't really have impact at all. Calling `produce` is effectively calling `poll` too, and since we have > 1 concurrency in prod, most of the polls will most likely be triggered by other calls to `produce`, not by this timer. The timer just sets the upper limit on waiting time for cases where produce call is not a rule but an exception, like producing to the error topic. But since these produce alls would be really rare, we can no care about 500 ms maximum delay here.

Pchelolo:
Let's deploy this separately from #134 since these both can have quite a big impact.

lib/kafka_factory.js
gwicke: I see, that sounds good for prod. I still think that it would be conceptually nicer to make reliable production fast enough for all normal uses, including tests. I don't see a strong reason why reliably producing to a single kafka broker would need to be much slower than a few dozen ms. But, agreed that it's not worth blocking on now.

Pchelolo:
Coverage fall is from all the code that should handle delivery errors/timeouts so we can't test those codepaths

Pchelolo:
So, can we maybe move this @gwicke @d00rman ?

PR #139, opened 2016-11-25
Rerender summary and mobile when wikidata item is undeleted

Pchelolo:
Need to rerender summary/mobile when wikidata item is undeleted to pull in the descriptions. Unfortunately, we can't do the same for deleted pages since CP doesn't have access to them after it was deleted and so we can't find the list of articles to purge.

Also, summary and mobile depend on the latest revision of the wikidata item and the latest one can't be restricted. So we don't need to rerender on revision-visibility-set event.

cc @wikimedia/services


Pchelolo:
Gerrit: https://gerrit.wikimedia.org/r/323633

PR #141, opened 2016-12-06
Ability to set per rule req timeouts and increased the default timeout

Pchelolo:
Bug: https://phabricator.wikimedia.org/T152073


gwicke:
:+1:

PR #142, opened 2016-12-06
Interleave continuation with leaf job processing

Pchelolo:
Per discussion in https://phabricator.wikimedia.org/T152557 we decided to allow doing whatever in internal CP topics, so it's OK to push 2 different kinds of messages to the transcludes topic.

Needs a config change to be deployed together or before the code.

This is ready for review and deployment. Also tested that the 'transition' mode will work just fine - we will not lose any jobs.

Bug: https://phabricator.wikimedia.org/T152229
cc @wikimedia/services

### DIFF
--- a/sys/dep_updates.js
+++ b/sys/dep_updates.js
@@ -6,6 +6,7 @@ const Template = HyperSwitch.Template;
 const utils = require('../lib/utils');
 const Title = require('mediawiki-title').Title;
 
+// TODO: needed only in transition period, to be removed
 const BACKLINKS_CONTINUE_TOPIC_NAME = 'change-prop.backlinks.continue';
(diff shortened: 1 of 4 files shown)


sys/dep_updates.js
```diff
@@ -25,8 +26,8 @@ function createBackLinksTemplate(options) {
             }
         })),
         getContinueToken: (res) => res.body.continue && res.body.continue.blcontinue,
-        continueTopic: BACKLINKS_CONTINUE_TOPIC_NAME,
         resourceChangeTags: [ 'backlinks' ],
+        leafTopicName: 'change-prop.backlinks.resource-change',
```
gwicke: (Minor/ later): Hard-coding such property names seems a bit ugly, but we can change this when we actually need to support different topics.

PR #143, opened 2016-12-07
Don't reuse request_id

Pchelolo:
Bug: https://phabricator.wikimedia.org/T151934


gwicke:
:+1: 

@Pchelolo, are there any reasons to not merge this right away?

Pchelolo:
Let's give @d00rman a chance to chip in?

gwicke:
@Pchelolo: Yup, otherwise happy to merge whenever you are ready.

d00rman:
I would honestly prefer setting another header or something similar. Sticking this in `x-triggered-by` with `, reqId:xxx` feels like a work-around IMHO and complicates parsing the field (unnecessarily IMO)

gwicke:
@d00rman, we need to build up a trace of request ids / topics etc, and x-triggered-by is designed to do exactly that, in a list of URLs. Since each subrequest will get a new request id, we will can have multiple `reqid:xxx` URLs in this trace.

gwicke:
@Pchelolo, this needs a rebase.

gwicke:
Did a quick & dirty rebase, lets see if the tests pass.

gwicke:
That clearly did not work.

@d00rman, if you are okay with the approach in principle I can look into doing a proper rebase.

d00rman:
I have soften my position on this so am ok with going with a different reqID for child events.

gwicke:
> I have softened my position on this so am ok with going with a different reqID for child events.

Okay, cool ;) Added a proper rebase to my todo.

gwicke:
Rebased in #203.

PR #146, opened 2016-12-08
Removed subscription to continue topics, not needed any more

Pchelolo:
Followup on #142


Pchelolo:
Checked that the lag in continue topic is 0 so it's safe to delete this code.

PR #147, opened 2016-12-09
Use shared kafka management tools

Pchelolo:
We would need to reuse those scripts in the trending service, so move them to their own package.


test/utils/clean_kafka.sh
```diff
@@ -1,44 +1,15 @@
...
-
-check ( ) {
-  PORT=$1
-  SERVICE_NAME=$2
-  if [ `nc localhost ${PORT} < /dev/null; echo $?` != 0 ]; then
-    echo "${SERVICE_NAME} not running, start it first with npm run start-kafka"
-    exit 1
-  fi
-}
+. node_modules/kafka-test-tools/clean_kafka.sh
```
d00rman: This assumes that the script is run from the root dir. Let's be precise:

```bash
. $(cd $(dirname $0) && pwd)/../../node_modules/kafka-test-tools/clean_kafka.sh
```

PR #152, opened 2016-12-15
Disable delivery reports and key

Pchelolo:
One more little step towards identifying the leak.
cc @wikimedia/services


gwicke:
:+1:, waiting for travis to concur.

PR #153, opened 2016-12-17
Switch back guaranteed producer

Pchelolo:
All bugs in the driver related to this has been fixed, so this is ready to go.

The memory leak wasn't found yet, so after deploying this we'd need to watch the memory really closely.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -40,14 +40,15 @@ class GuaranteedProducer extends kafka.Producer {
         super(conf, topicConf);
         this._pending = {};
 
-        this.on('delivery-report', (report) => {
+        this.on('delivery-report', (err, report) => {
             const reportKey = `${report.topic}:${report.key}`;
-            const resolve = this._pending[reportKey];
-            if (resolve) {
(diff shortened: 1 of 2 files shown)


lib/kafka_factory.js
```diff
@@ -24,39 +24,35 @@ const CONSUMER_TOPIC_DEFAULTS = {
...
         this.on('delivery-report', (report) => {
             const reportKey = `${report.topic}:${report.key}`;
             const resolve = this._pending[reportKey];
-            if (!resolve) {
-                this._logger.log('error/produce', {
-                    message: 'Unknown resolver in delivery report',
-                    report
-                });
-                return;
+            if (resolve) {
```
gwicke: If the timeout elapses, won't this closure here still be registered with the driver & thus hold onto data it closed over?

Pchelolo: This closure is created once and not on each message, and then called many times. So it doesn't matter here

Pchelolo:
The vacation times are over, let's get this moving?

PR #154, opened 2016-12-19
Move transclusion updates to the page_edit rule

d00rman:
Currently, figuring out if a page is transcluded is a separate rule from
page_edit. This means that the transclude process will run even if there
are failures in rendering the new revision in RB/Parsoid. The behaviour
is not ideal in such cases, especially if rendering the original page is
blacklisted in RESTBase. An extreme case has been discovered this
week-end where a user page got edited repeatedly by bot, and as it
happens, that page is transcluded in thousands of other user pages.

By merging these two actions into the page_edit rule we ensure that the
transclusion process will be started only if the initial render
succeeded. This has two important consequences:
- it makes CP respect RESTBase's blacklists (as such requests will not
  be retried)
- it makes CP less prone to overload the system, since the transclusion
  process will not start if RESTBase returns a 5xx for whatever reason

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -178,22 +178,29 @@ spec: &spec
                   status:
                     - '5xx'
                     - 404 # Sometimes occasional 404s happen because of the mysql replication lag, so retry
+                # Test-only. We use undefined rev_parent_id to test backlinks so we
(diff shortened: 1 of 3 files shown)


d00rman:
cc @wikimedia/services

Pchelolo:
Great idea @d00rman. I'll wait for travis and merge

PR #155, opened 2016-12-23
Start testing with kafka 0.10 as well

Pchelolo:
Kafka 0.10 will bring us a lot of highly needed features, like support for timestamp index and seeking to the timestamp. Obviously, it's too new to update the cluster, and all the nice features are not supported by `librdkafka` of the driver yet, but let's start testing the newer version in travis sooner to build up the confidence when the time for an upgrade comes.

cc @wikimedia/services


gwicke:
Didn't know about the build-per-env-line feature in travis. Neat!

d00rman:
:100:

PR #160, opened 2017-01-31
Propagate wikidata description change on undo/revert action

mdholloway:
We need to add 'undo' to the regex for detecting Wikidata description
edits so that this type of change event also gets propagated.


Pchelolo:
LGTM, waiting for tests and deploying. Thank for providing a special test for it, honestly I'd be too lazy to write a tests here :)

Pchelolo:
Created a corresponding commit in the deploy repo: https://gerrit.wikimedia.org/r/335294

PR #161, opened 2017-02-06
Disable CP precaching in Wikis where ORES review tool is enabled

Ladsgroup:
It won't be needed since the extension does the precaching too

Bug: T157206


d00rman:
Hm, so the extension is using the job queue now for that @Ladsgroup ?

Ladsgroup:
It has been since the beginning (https://grafana.wikimedia.org/dashboard/db/ores-extension?from=now-1h&to=now) but it's only for damaging model and not goodfaith or reverted.

PR #162, opened 2017-02-10
Make CP use minfied responses from ORES

Ladsgroup:
See https://phabricator.wikimedia.org/T157693


Pchelolo:
Test is failing because the request is not matching the expectation any more: https://github.com/wikimedia/change-propagation/blob/master/test/feature/update_rules.js#L443-L447

PR #165, opened 2017-02-28
Prototype of the idea of reporting metrics for RESTBase varnish

Pchelolo:
In order to better isolate, we can run this metrics reporting code in a separate instance of change-prop, thus the new config file. 

Added support for non-prefixed topic names, switching off retries whatsoever. The metrics module is caching the routers, and if the specs on multiple domains are the same, the routers are reused. Sampling 1% by default, but that's configurable. 

cc @wikimedia/services

### DIFF
--- a/config.example.metrics.yaml
+++ b/config.example.metrics.yaml
@@ -1,0 +1,58 @@
+spec: &spec
+  x-sub-request-filters:
+    - type: default
+      name: http
+      options:
+        allow:
+          - pattern: /^https?:\/\//
+            forward_headers:
+              user-agent: true
+  title: The Change Propagation metrics reporter root
+  paths:
+    /sys/metrics:
+      x-modules:
+        - path: sys/metrics.js
+          options:
+            templates:
+              spec_request:
+                  method: 'get'
+                  uri: 'https://{{message.uri_host}}/api/rest_v1/?spec'
+    /sys/queue:
+      x-modules:
+        - path: sys/kafka.js
+          options:
+            metadata_broker_list: 127.0.0.1:9092
+            dc_name: '' # VarnishKafka topics are not prefixed with a DC name
+            enable_retries: false
+            startup_delay: 0
+            consumer:
+              # These options should not be copied to puppet config.
+              # We're using this config for testing, so need to configure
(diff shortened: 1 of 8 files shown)


config.example.metrics.yaml
```diff
@@ -0,0 +1,58 @@
...
+              # for minimal latency
+              fetch.wait.max.ms: "1"
+              fetch.min.bytes: "1"
+              queue.buffering.max.ms: "1"
+            concurrency: 250
+            templates:
+              varnish_metrics_collect:
+                topic: webrequest_text
+                match:
+                  uri_path: '/^\/api\/rest_v1/'
```
gwicke: Ahh, nice that we can just use the built-in matching functionality.

sys/metrics.js
```diff
@@ -0,0 +1,90 @@
...
+                }
+                return router;
+            });
+        }
+        return this._router_cache.get(domain);
+    }
+
+    webrequest(hyper, req) {
+        const message = req.body;
+        return this._getRouter(hyper, message)
```
gwicke: Minor: It might be slightly more efficient to synchronously check the cache first, and only introduce the promise when needed.

sys/metrics.js
```diff
@@ -0,0 +1,90 @@
...
+        .then((router) => {
+            const handler = router.route(message.uri_path.replace(/^\/api\/rest_v1/, ''));
+            const path = handler.value.path;
+            let statName = `${path}.${message.http_method.toUpperCase()}.`;
```
gwicke: Also strip the domain as in https://github.com/wikimedia/hyperswitch/blob/master/lib/filters/metrics.js#L9 ?

Pchelolo: This is the public path, so the domain is not in there fro the beginning.

gwicke: Ah, makes sense. Thanks for the clarification!

sys/metrics.js
Pchelolo: Normally the promise is already resolved, so we're not creating a new promise here but reusing the old one instead.

Pchelolo:
I believe we will not do this any more.

PR #166, opened 2017-03-03
WIP: Added support for sampling the rule execution

Pchelolo:


config.example.wikimedia.yaml
```diff
@@ -182,6 +182,9 @@ spec: &spec
                   status:
                     - '5xx'
                     - 404 # Sometimes occasional 404s happen because of the mysql replication lag, so retry
+                sample:
```
Pchelolo: Don't think that's needed in the example, let's better add a section in `README`

lib/rule.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+/* eslint no-bitwise: ["error", { "allow": ["~", "<<"] }] */
+/* jshint bitwise: false */
```
Pchelolo: Put this on the method-level perhaps?

lib/rule.js
```diff
@@ -10,6 +14,68 @@ const DEFAULT_RETRY_LIMIT = 2;      // At most two retries
 // Set a larger request timeout then RESTBase default 6 minutes;
 const DEFAULT_REQUEST_TIMEOUT = 7 * 60 * 1000;
 
+class Sampler {
```
Pchelolo: Sampler is quite big already, perhaps it deserves getting it's own file?

lib/rule.js
```diff
@@ -10,6 +14,68 @@ const DEFAULT_RETRY_LIMIT = 2;      // At most two retries
...
+    /**
+     * Returns true if this request should be sampled, false if it should
+     * be ignored.
+     *
+     * @param   {Object} context
+     * @returns {boolean}
+     */
+    accept(context) {
+        const hashSource = this._hashSourceTemplate.expand(context);
+        return (Sampler.hash(hashSource) <= this._tokens[this._threshold]);
```
Pchelolo: `this._tokens[this._threshold]` is always the same, maybe calculate it upfront in the constructor? Avoids the need to store the tokens array as well. Then the tokens array could be stored in some const on the file level and only calculated once when the file is `require`d

PR #170, opened 2017-03-15
Created autopolling producer

Pchelolo:
Better encapsulation of `poll` call and better promise rejection handling. Also for testing decreased the side of the buffer - if poll is not called correctly tests would start failing.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -36,16 +36,16 @@ class GuaranteedProducer extends kafka.Producer {
     /**
      * @inheritDoc
      */
-    constructor(conf, topicConf, logger) {
+    constructor(conf, topicConf, log) {
         super(conf, topicConf);
-        this._logger = logger;
+        this._log = log;
         this._pending = {};
 
         this.on('delivery-report', (report) => {
             const reportKey = `${report.topic}:${report.key}`;
             const resolve = this._pending[reportKey];
             if (!resolve) {
-                this._logger.log('error/produce', {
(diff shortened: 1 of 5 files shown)


sys/kafka.js
```diff
@@ -93,9 +93,8 @@ class Kafka {
 
             // TODO: This has to return a promise for guaranteed producer.
             // TODO: This is a quick fix to be rewritten.
-            this.producer.produce(`${this.kafkaFactory.produceDC}.${message.meta.topic}`, 0,
+            return this.producer.produce(`${this.kafkaFactory.produceDC}.${message.meta.topic}`, 0,
```
d00rman: The above TODOs can now be removed, couldn't they?

lib/kafka_factory.js
```diff
@@ -107,6 +107,34 @@ class GuaranteedProducer extends kafka.Producer {
...
+            try {
+                const result = super.produce(topic, partition, message, key);
+                if (result !== true) {
+                    return reject(result);
+                }
+                return resolve(result);
+            } catch (e) {
+                return reject(e);
+            } finally {
+                this.poll();
```
d00rman: Are we sure that this will be called before the returned Promise is being activated?

Pchelolo: Yep, just checked that.

sys/kafka.js
Pchelolo: Oh, sorry, forgot about it!

PR #171, opened 2017-03-16
Switch to guaranteed producer using opaque

Pchelolo:
Attempt number 3.

The newest version of the kafka driver support a message `opaque` feature. In C++ land it's basically attaching an arbitrary pointer to the message and then returns the pointer together with a delivery report. Node driver supports providing an arbitrary Javascript object, pointer to which would be used as an opaque in C++ land, so we can get back the same exact object. Since the object in JS land gets out of the scope and might be GCed normally, driver uses native `Nan::Persistent` in order to hold on to the object while it's in the C++ land and resets the persistent before passing it back to JS land.

Using this feature we now can pass promise `resolve` and `reject` functions through native code, get them back in the delivery report and call them which makes guaranteed producer implementation so much easier (and scarier)

I've verified the driver code around this new feature we're adopting and done very extensive memory-leaks testing - locally all runs fine. Doesn't guarantee we won't have a memory leak in production again though..

Also, this version doesn't `poll` on each and every produce call, but polls once in 100 ms instead. It is a bit riskier, but the buffer size is 100.000 messages and I don't think we'd ever produce fast enough to overflow the buffer.

cc @wikimedia/services

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -48,7 +48,7 @@ class GuaranteedProducer extends kafka.Producer {
         });
 
         this.on('ready', () => {
-            this._pollInterval = setInterval(() => this.poll(), 100);
+            this._pollInterval = setInterval(() => this.poll(), 10);
         });
     }
 
@@ -72,6 +72,8 @@ class GuaranteedProducer extends kafka.Producer {
                 process.nextTick(() => {
                     reject(e);
                 });
+            } finally {
+                this.poll();
             }
         });
     }
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -24,11 +24,11 @@ const CONSUMER_TOPIC_DEFAULTS = {
 };
 
 const PRODUCER_DEFAULTS = {
-    // dr_cb: true
+    dr_cb: true
 };
 
 const PRODUCER_TOPIC_DEFAULTS = {
-    // 'request.required.acks': 1
+    'request.required.acks': 1
 };
 
 /*eslint-disable */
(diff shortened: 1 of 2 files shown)


lib/kafka_factory.js
```diff
@@ -24,72 +24,52 @@ const CONSUMER_TOPIC_DEFAULTS = {
...
             }
-
-            delete this._pending[reportKey];
-            resolve(report);
+            return reporter.resolver(report);
         });
 
-        this._pollInterval = setInterval(() => this.poll(), 500);
+        this.on('ready', () => {
+            this._pollInterval = setInterval(() => this.poll(), 100);
```
gwicke: How long can a poll call take? Is there any chance it would ever take longer than the interval (100ms)?

Pchelolo: @gwicke `poll` is asynchronous as far as I understand

gwicke:
This looks good to me. On the C++ side, I think [this is the section where the void* is set up](https://github.com/Blizzard/node-rdkafka/blob/05ea3def4034fd7eb6981709abd064a911c0eae6/src/producer.cc#L389-L396).

Pchelolo:
@gwicke Yep, that's the code. Not guaranteed it won't leak again, local testing didn't show the leak before too, but let's hope for the best :)

gwicke:
The corresponding [dereferencing is happening here](https://github.com/Blizzard/node-rdkafka/blob/ef49abe651ab02c46128b3284a03ed93bbe5daaa/src/callbacks.cc#L279-L290). The Nan::Persistent handling looks correct to me, but I haven't dug very deep to verify that all (error-) code paths really hit that callback handling code. But, we'll find out soon enough I guess ;)

Pchelolo:
Tested for native leaks using `instruments` (`gdb` Mac alternative) and nothing found..

Pchelolo:
Done some more testing and update the PR to call `poll` after every call to `produce` for safety. We can consider making less polls in the future, but we should start conservatively.

lib/kafka_factory.js
```diff
@@ -24,74 +24,56 @@ const CONSUMER_TOPIC_DEFAULTS = {
...
                         reject(result);
                     });
                 }
             } catch (e) {
                 process.nextTick(() => {
-                    delete this._pending[reportKey];
                     reject(e);
                 });
+            } finally {
+                this.poll();
```
d00rman: Why do we need to poll here if the producer polls every 10ms?

Pchelolo: @d00rman When we produce that 10ms sets the low bar for a single production latency. In practice that shouldn't be a problem, but since we use a single producer for all the production, I've decided to start from a very high-frequency polling to avoid any possibility of message buffer overflow. When we see this works we can optimize and poll less frequently, but in the beginning I wanted to be very safe.

PR #172, opened 2017-03-22
Support regexes topics and refactor subscription process

Pchelolo:
A little pet project of mine finally got working.

This idea is to support regexes in topic property of the rule definition for rules that should listen to different topics but do exactly the same work. Right now we can use them for the `transcodes` topics - the ones we're using to offload summary/mobile defenders generated by template updates and priorities re renders that we're doing for normal edits. Right now it's not super-required feature, but it's a step towards the JobQueue yearly goal and deploying this early can give us a good change to test it and debug this piece before hand.

Also, a little refactoring to abstract out the notion of a subscription which contains of 2 executors (normal and retry) for normal rules and many executors for regexes rules.

cc @wikimedia/services 
Bug: https://phabricator.wikimedia.org/T157093

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -46,7 +46,7 @@ spec: &spec
             templates:
 
               summary_definition_rerender: &summary_definition_rerender_spec
-                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
+                topic: '/^(?:change-prop\.transcludes\.)?resource[-_]change$/'
                 retry_limit: 2
                 retry_delay: 500
                 retry_on:
@@ -89,7 +89,7 @@ spec: &spec
                         cache-control: no-cache
 
               mobile_rerender: &mobile_rerender_spec
-                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
+                topic: '/^(?:change-prop\.transcludes\.)?resource[-_]change$/'
                 retry_limit: 2
                 retry_delay: 500
                 retry_on:
@@ -112,7 +112,7 @@ spec: &spec
                     cache-control: no-cache
 
               purge_varnish: &purge_varnish_spec
-                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
+                topic: '/^(?:change-prop\.transcludes\.)?resource[-_]change$/'
                 match:
                   meta:
                     uri: '/^https?:\/\/[^\/]+\/api\/rest_v1\/(?<title>.+)$/'
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -46,7 +46,7 @@ spec: &spec
             templates:
 
               summary_definition_rerender: &summary_definition_rerender_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
                 retry_limit: 2
                 retry_delay: 500
                 retry_on:
@@ -88,12 +88,8 @@ spec: &spec
                       headers:
                         cache-control: no-cache
 
-              summary_definition_rerender_transcludes:
-                <<: *summary_definition_rerender_spec
-                topic: change-prop.transcludes.resource-change
-
               mobile_rerender: &mobile_rerender_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
                 retry_limit: 2
                 retry_delay: 500
                 retry_on:
@@ -115,12 +111,8 @@ spec: &spec
                   headers:
                     cache-control: no-cache
 
-              mobile_rerender_transcludes:
-                <<: *mobile_rerender_spec
-                topic: change-prop.transcludes.resource-change
-
               purge_varnish: &purge_varnish_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
                 match:
                   meta:
                     uri: '/^https?:\/\/[^\/]+\/api\/rest_v1\/(?<title>.+)$/'
@@ -132,10 +124,6 @@ spec: &spec
                   body:
                     - meta:
                         uri: '//{{message.meta.domain}}/api/rest_v1/{{match.meta.uri.title}}'
-
-              purge_varnish_transcludes:
-                <<: *purge_varnish_spec
-                topic: change-prop.transcludes.resource-change
 
               # RESTBase update jobs
               mw_purge:
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -149,6 +149,7 @@ class MetadataWatch extends EventEmitter {
     }
 
     getTopics() {
+        const dcRemoveRegex = new RegExp(`^${this._consumeDC}\\.`);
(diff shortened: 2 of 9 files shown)


config.example.wikimedia.yaml
```diff
@@ -46,7 +46,7 @@ spec: &spec
             templates:
 
               summary_definition_rerender: &summary_definition_rerender_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
```
d00rman: Did you mean(TM) `/^(?:change-prop`... ?

config.example.wikimedia.yaml
```diff
@@ -88,12 +88,8 @@ spec: &spec
                       headers:
                         cache-control: no-cache
 
-              summary_definition_rerender_transcludes:
-                <<: *summary_definition_rerender_spec
-                topic: change-prop.transcludes.resource-change
-
               mobile_rerender: &mobile_rerender_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
```
d00rman: Idem

config.example.wikimedia.yaml
```diff
@@ -115,12 +111,8 @@ spec: &spec
                   headers:
                     cache-control: no-cache
 
-              mobile_rerender_transcludes:
-                <<: *mobile_rerender_spec
-                topic: change-prop.transcludes.resource-change
-
               purge_varnish: &purge_varnish_spec
-                topic: resource_change
+                topic: '/^(:?change-prop\.transcludes\.)?resource[-_]change$/'
```
d00rman: Yup, idem

config.example.wikimedia.yaml
Pchelolo: Heh.. How the hell does it work? :)

lib/kafka_factory.js
```diff
@@ -119,6 +118,62 @@ class AutopollingProducer extends kafka.Producer {
...
+        return new P((resolve, reject) => {
+            this._consumer.getMetadata(undefined, (err, res) => {
+                if (err) {
+                    return reject(err);
+                }
+                resolve(res.topics.filter((topicInfo) =>
+                    topicInfo.name.startsWith(this._consumeDC))
+                .map((topicInfo) => {
+                    if (this._consumeDC) {
+                        return topicInfo.name.replace(`${this._consumeDC}.`, '');
```
d00rman: Let's make this a regexp and anchor it to beginning of the topic name. As it currently is, if by any chance a topic include the DC as a legit part of the name, it would get removed.

lib/rule.js
```diff
@@ -135,10 +135,18 @@ class Rule {
...
         this.spec = spec || {};
 
-        this.topic = this.spec.topic;
-        if (!this.topic) {
+        const  topic = this.spec.topic;
+        if (!topic || typeof topic !== 'string' || !topic.length) {
             throw new Error(`No topic specified for rule ${this.name}`);
         }
+
+        if (/^\/.+\//.test(topic)) {
```
d00rman: `/^\.+\/$/`

lib/rule_subscriber.js
```diff
@@ -0,0 +1,154 @@
...
+        });
+
+        this._subscribed = false;
+        this._executors = [];
+    }
+
+    _subscribeTopic(topicName) {
+        const topicRule = this._rule.clone(topicName);
+
+        this._log('info/subscription', {
```
d00rman: There are two entries logged for a regex subscription, one following the other (the first one being on L56). Should probably remove one of the two.

d00rman: The one on L56 should go probably.

Pchelolo: Nope, there's not. This one would be called for all the topics the exist at the time CP is started. The other one (L56) will be logged when a new topic is created and we detect it and subscribe

lib/rule_subscriber.js
```diff
@@ -0,0 +1,154 @@
...
+
+    unsubscribe() {
+        if (this._subscribed) {
+            this._executor.close();
+            this._retryExecutor.close();
+        }
+    }
+}
+
+class RegexTopicSubscription {
```
d00rman: Should this class extend `BasicSubscription` ?

lib/rule_subscriber.js
```diff
@@ -0,0 +1,154 @@
...
+    constructor(options, kafkaFactory) {
+        this._kafkaFactory = kafkaFactory;
+        this._options = options;
+
+        this._subscriptions = [];
+        this._metadataWatch = undefined;
+    }
+
+    _createSubscription(hyper, rule) {
+        if (rule.isRegexRule()) {
```
d00rman: Nit: imho the method would be more readable if you put `if (!this.rule.isregexRule())` first as the block is much shorter.

lib/rule_subscriber.js
Pchelolo: Why? It doesn't have anything in common implementation-vise. In Java world they would share a common interface, but unfortunately we don't have interfaces here.. We could make a common superclass `AbsctractSubscription` though..

d00rman: Exactly because they are expected to have the same interface. But let's not go that far as to create futile Java-esque structures :)

Pchelolo: @d00rman You don't even imagine how much did I want to create a `SubscriptionFactory`...

lib/rule_subscriber.js
d00rman: Wait, L56 triggers a log entry. Then, L60 calls this method and the first thing it does is to log it again. It seems to me L56 is superfluous.

lib/rule_subscriber.js
d00rman: loool

lib/rule_subscriber.js
Pchelolo: Oh damn! Right... That wasn't the intention..

lib/rule.js
Pchelolo: Why? We want to detect stuff that starts with `/` and ends with `/`..

Pchelolo:
@d00rman Addressed your comments. What do you think about the idea in general?

d00rman:
> What do you think about the idea in general?

Really solid. Really nice work! I was worried about tying up one extra libuv thread per regex subscription, but this matters only in the edge case where one regex manages to subscribe to only one topic, which shouldn't happen in practice (if we are careful, of course).

Pchelolo:
> Really nice work! I was worried about tying up one extra libuv thread per regex subscription, but this matters only in the edge case where one regex manages to subscribe to only one topic, which shouldn't happen in practice (if we are careful, of course).

You mean because of the metadata refresher thing? Cool thing is that it doesn't tie up a thread. Only consuming ties up a whole thread, fetching metadata doesn't! And also, it's a singleton created lazily and it's global for the whole subscriber, that means there's only one refresher per worker.

d00rman:
Oh, indeed! Wooow! Bravo :clap:

Pchelolo:
@d00rman Do you think we can move on this? I'd like to deploy it in 2 stages, first just deploy the code without the config change to verify it's not breaking anything existing, and then move on the config change. I think that would minimize the risk

PR #177, opened 2017-03-31
Update driver to 0.10

Pchelolo:
1. Update driver to 0.10, adapt to the new API
2. After release of new eslint it enforces some rules more properly - fix that up.

cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,9 @@
     "preq": "^0.5.2",
     "ajv": "^4.11.5",
     "mocha-eslint":"^3.0.1",
-    "eslint-config-node-services": "^2.0.2",
+    "eslint-config-node-services": "^2.1.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -44,23 +44,23 @@
...
+    "coveralls": "^2.13.0",
+    "js-yaml": "^3.8.2",
+    "nock": "^9.0.11",
     "preq": "^0.5.2",
-    "ajv": "^4.11.3",
+    "ajv": "^4.11.5",
     "mocha-eslint":"^3.0.1",
-    "eslint-config-node-services": "^1.0.6",
-    "kafka-test-tools": "^0.1.0"
+    "eslint-config-node-services": "^2.0.2",
```
d00rman: Need to add eslint and eslint-plugin-json

PR #178, opened 2017-04-05
General-purpose rate-limiting with Redis

Pchelolo:
Basically, there will be a limiter module. In it's config you can specify limiters with arbitrary names, for example `blacklist`. Then in each rule you can add `limiters` stanza to specify a template of the key that would be used in a particular limiter. The limiter names will be predefined, because the limiter counters needs to be updated under different circumstances. For example for the `blacklist` limiter we have to bump the counter in case of a 503 and retry limit is exceeded, while for `rerender_rate` limiter we have to bump the counter on every successful rerender of a certain revision. I'm not quite sure how to generalize it even more and whether we need to generalize it more..

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -50,6 +50,8 @@ class BaseExecutor {
         this._pendingMsgs = [];
         this._toCommit = undefined;
         this._consuming = false;
+
+        this._rateLimiters = Object.assign(this.options.limiters || {}, rule.spec.limiters || {});
     }
 
     subscribe() {
@@ -98,7 +100,12 @@ class BaseExecutor {
                 // Note: we don't return the promise here since we wanna process messages
                 // asynchronously from consuming them to be able to fill up the pendingMsg
                 // queue and achieve the level of concurrency we want.
-                this.processMessage(message, handler)
+                this._rateLimitMessage(message)
+                .then((isRateLimited) => {
+                    if (!isRateLimited) {
+                        return this.processMessage(message, handler);
+                    }
+                })
                 .finally(() => {
                     this._notifyFinished(msg);
                     if (this._pendingMsgs.length < this.concurrency && !this._consuming) {
@@ -128,6 +135,28 @@ class BaseExecutor {
             } else {
                 this._consuming = false;
             }
+        });
+    }
+
+    /**
+     * Checks whether a message should be rate-limited
+     * @param {Object} message the message to check
+     * @return {boolean}
+     * @private
+     */
+    _rateLimitMessage(message) {
+        return P.all(this.rule.getRateLimiterTypes().map((type) => {
+            const key = this.rule.getLimiterKey(type, message);
+            return this.hyper.get({
+                uri: `/sys/limit/${type}/${key}`
+            });
+        }))
+        .thenReturn(false)
+        // Will throw if any of the limiters failed,
+        // the error message will say which one is failed.
+        .catch({ status: 429 }, (e) => {
+            // TODO: log
+            return true;
         });
     }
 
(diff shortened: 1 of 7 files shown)


sys/rate_limiter.js
```diff
@@ -0,0 +1,86 @@
...
+        this._log = this._options.log || (() => {});
+
+        // TODO: verify and pass in the redis options
+        this._client = redis.createClient(this._options.redis);
+
+        // TODO: configurable options
+
+        this._LIMITERS = new Map();
+        Object.keys(this._options.limiters).forEach((type) => {
+            const limiterOpts = this._options.limiters[type];
```
d00rman: It would be good to either enforce the existence of the `interval` and `limit` keys or give defaults for them so that the structure is known

sys/rate_limiter.js
```diff
@@ -0,0 +1,86 @@
...
+            this._LIMITERS.set(type, new Limiter(this._client, [
+                limiterOpts
+            ], { prefix: `CPLimiter_${type}` }));
+        });
+    }
+
+    _execLimiterFun(fun, hyper, type, key) {
+        const limiter = this._LIMITERS.get(type);
+        if (!limiter) {
+            return { status: 204 };
```
d00rman: It would be good to emit a warning that the desired type hasn't been registered as that should not happen.

lib/rule.js
```diff
@@ -163,6 +163,13 @@ class Rule {
         this.shouldRetry = _compileErrorCheckCondition(this.spec.retry_on);
         this.shouldIgnoreError = _compileErrorCheckCondition(this.spec.ignore);
 
+        this._limiterKeyTemplates = {};
+        if (this.spec.limiters) {
+            Object.keys(this.spec.limiters).forEach((type) => {
+                this._limiterKeyTemplates[type] = new Template(this.spec.limiters[type]);
```
d00rman: Idem, we should probably check a template has actually been given.

Pchelolo:
@d00rman hehe, thank you for the review, but I didn't think you would look into the details of this, it's a very early-stage WIP still

lib/rule.js
Pchelolo: If the template is wrong the compilation will fail and service will fail starting.

Pchelolo:
Still not completely ready for review. We've decided to reiplement it in a more generic way. Stay toned.

Pchelolo:
This is a step 1 of the current idea.

The whole idea is that on each rule we'd be able to report to any number of limiters with a special 'cost' function. Currently, as a first iteration, the 'cost' function is not configurable and it's just '1' if the request failed or timed out.

Pchelolo:
@gwicke @d00rman This is ready for review. This is the first, non-functional version - it doesn't block anything, just logs when something should be blocked. 

I would like to move this one to prod before going further on this (refactoring to a limiting filter, adding more complex cost functions etc) to test that everything works properly on the Redis side. 

So, this one is ready for a review.

lib/base_executor.js
```diff
@@ -236,32 +289,37 @@ class BaseExecutor {
...
+                return { status: 200 };
+            }
+            return P.each(handler.exec, (tpl) => {
+                const request = tpl.expand(expander);
+                request.headers = Object.assign(request.headers, {
+                    'x-request-id': origEvent.meta.request_id,
+                    'x-triggered-by': utils.triggeredBy(retryEvent || origEvent)
+                });
+                return this.hyper.request(request)
+                    .tap(redirectCheck)
```
d00rman: Nit: wrong indentation

lib/rule.js
d00rman: Right, but with sooo many config options, having a user-friendly error would be much appreciated by users :)

d00rman:
Some nits, otherwise :100:

Pchelolo:
@d00rman Done!

PR #179, opened 2017-04-12
Added ability to run executor steps in parallel

Pchelolo:
For ORES it doesn't make sense to run the reqs to both Das sequentially, so support doing parallel requests. 

I guess at some point we might wanna bring in the full-blown tempting support from hyper switch here, but for now this simple change is enough.

cc @wikimedia/services

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ rules to workers equally.
 
 The rule can contain the following properties:
 - **topic** A name of the topic to subscribe to.
+- **parallel** Whether to run the executor steps sequentially or in parallel. Default: `false` 
 - **match** An optional predicate for a message. The rule is executed only if all of the `match`
 properties were satisfied by the message. Properties could be nested objects, constants
 or a regex. Regex could contain capture groups and captured values will later be accessible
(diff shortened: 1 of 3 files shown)


gwicke:
Actually, I'm not entirely sure that it makes sense to execute both updates in the same job. Won't this couple the update speed between the two DCs? Wouldn't one DC being down  / slow basically stop updates for the healthy DC?

Pchelolo:
P.all doesn't get cancelled in case one of them fails, but actually you're making a good point. I guess having `ores_eqiad` and `ores_codfw` separate rules would be indeed be better. We didn't quite think that through.

I'll prepare the update to the deploy repo. This PR is not needed in that case, feel free to close it.

Pchelolo:
Hm.. Actually, separating the rules properly would require a puppet change (we'd wanna convert the list of ORES Uris to a map `dc_name` => `ores_uri` for proper rule naming), and since this solution is temporary (until the new recache endpoint is there) I think we can go with this feature for now and implement a more proper solution with the switch to `/precache` endpoint.

Pchelolo:
Not needed any more after https://gerrit.wikimedia.org/r/#/c/347980/

PR #180, opened 2017-05-10
Sampler: Bug Fix: Return a Promise instead of an Object

d00rman:
Change Propagation's rule execution mechanism expects a Promise to be
returned when executing a rule. Alas, if the sampler is active, only an
Object is being returned. This commit fixes the obvious error.

This is a follow-up on PR #166


d00rman:
The Travis failure is unrelated (only for Node 6, missing an eslint module)

PR #181, opened 2017-05-17
Bug Fix: Test mode: Remove the processed message from the queue

d00rman:
Regardless of whether we are going to actually commit the offset or not,
if a message has been processed it needs to be removed from the list of
pending messages, as otherwise a deadlock occurs because Change
Propagation will not process new messages while the length of that array
is greater than the allowed concurrency.

This is another follow-up on #166


eevans:
Nice catch!

PR #193, opened 2017-06-25
Disable precachign for wikis that have ORES review tool enabled

Ladsgroup:
And add new wikis

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -369,18 +369,6 @@ spec: &spec
                         format: 'json'
                   - match:
                       meta:
-                        domain: cs.wikipedia.org
-                      performer:
-                        user_is_bot: false
-                    exec:
-                      uri: 'https://ores.wikimedia.org/v2/scores/cswiki/'
-                      query:
-                        models: 'reverted'
-                        revids: '{{message.rev_id}}'
-                        precache: true
-                        format: 'json'
-                  - match:
-                      meta:
                         domain: en.wiktionary.org
                       performer:
                         user_is_bot: false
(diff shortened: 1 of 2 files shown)


Ladsgroup:
https://gerrit.wikimedia.org/r/#/c/361300/

Pchelolo:
@Ladsgroup Out of curiosity - what is the review tool and how does it affect precaching?

Ladsgroup:
ORES review tool is the extension that is enabled on wikis that have damaging and goodfaith models. It triggers a job after each edit (non bot edits and depends on the config of the wiki, so for example it only triggers the job after edits in main namespace for wikidata because it's useless in other namespaces). It's practically precaching but we had to keep both since it didn't have goodfaith model, it has it now and thus these jobs are redundant.

d00rman:
@Ladsgroup q: why is precaching not consolidated in one place? Having precaching updates here for some wikis and using the JobQueue for others is a little bit confusing. What is the benefit in having the review tool doing it as opposed to having it all being emitted by CP?

Ladsgroup:
The main benefit is that beside everything that CP does, it also stores it in ores_classification table so it's usable by lots of features directly (mediawiki API, ORES review tool, new RC filters, etc.) Without these jobs, none of these features were possible.

Pchelolo:
@Ladsgroup Hm.. If we look at this in a broader perspective into the future, when we start emitting the special `revision-score` event and switch this to a `/precache` endpoint, CP won't have this giant config specific to the wiki - we would simply call ORES for every revision create to emit the new event. 

If CP is faster - it will act as a precache and the the job will just use the result, if the job is faster - CP will just use the cached version. Since you merge parallel requests in ORES, this wouldn't create double load, serving the scores from the cache should have close-to-zero cost.

d00rman:
I agree with @Pchelolo, consolidating the ways precaching is achieved is the way to go, both from an architectural as well as from a conceptual point of view. Regarding the storing of results, this can be greatly simplified if RESTBase is use to store those, as it can act as both storage and cache for the data, in this way eliminating the need for Redis and MW/MariaDB.

For the time being, though, I will merge this (with an unhappy face) in order not to block progress.

PR #194, opened 2017-06-26
Revert "Generalized the deduplication as a filter"

Pchelolo:
Reverts wikimedia/change-propagation#192

I would like to take this back. When I started to create a redid-based filter I've realized that this is a half-baked solution. WE can actually make this way more generic and apply to all the rules by default if we make some refactoring within the change-prop codebase.

cc @wikimedia/services


gwicke:
LGTM.

PR #195, opened 2017-06-30
Only update when the content of the revision actually changed

Pchelolo:
Some actions on the wiki, like page protection changes, can create a new revision without actually changing the page contents. Obviously, It doesn't make sense to rerender a page if it's content hasn't changed.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -193,17 +193,19 @@ spec: &spec
                   status:
                     - '5xx'
                     - 404 # Sometimes occasional 404s happen because of the mysql replication lag, so retry
+                match:
+                  rev_content_changed: true
+                match_not:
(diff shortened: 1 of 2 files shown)


gwicke:
Will manual null edits and action=purge calls still cause a re-render?

Pchelolo:
@gwicke null edits and purges are separate rules and separate events.

Pchelolo:
Deploy repo: https://gerrit.wikimedia.org/r/#/c/362453/1

gwicke:
> @gwicke null edits and purges are separate rules and separate events.

Okay, great!

LGTM.

Pchelolo:
Perfecto, next Thursday we can deploy this. I will help with the ceb.wiki problem

Pchelolo:
The changes in MW were deployed, so this is unblocked now and can go!

PR #196, opened 2017-07-06
Extend blacklisting for derived updates

Pchelolo:
After a day of trying to figure out the core reason of the https://phabricator.wikimedia.org/T169911 with no success, we can at least enable blacklisting for all of the derived content types. 

I'm changing the key for the HTML blacklist, so we will loose the blacklisting that's already been done, but it's ok, there's not too much stuff in there anyway.

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -71,6 +71,8 @@ spec: &spec
                 retry_on:
                   status:
                     - '5xx'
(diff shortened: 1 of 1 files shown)


gwicke:
:+1:

Pchelolo:
Gerrit: https://gerrit.wikimedia.org/r/363731

PR #197, opened 2017-07-25
Make a generic redis-based deduplicator

Pchelolo:
So we're gonna take the same approach to deduplication as we do for blacklisting.

We will have a separate deduplicator module with a configured redistribution client and will unconditionally send all the messages to the deduplicator. Then it will, just like MediaWiki version, first check whether the message individually is a duplicate - e.g. if the message with exact same ID was recently processed by exact same rule. Also, if there's information about the `root_event` - attempt to deduplicate based on the root event.

This also switches the dependency processor to the new reduplication method. Will need to update schemas, but it doesn't really matter, because these events are not going via the proxy service, so the schemas are just for the documentation purposes.

@wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -13,7 +13,7 @@ spec: &spec
       x-modules:
         - path: sys/rate_limiter.js
           options:
-            redis:
+            redis: &redis_config
               host: localhost
               port: 6379
 
@@ -28,6 +28,11 @@ spec: &spec
                 # errors per hour
                 - interval: 3600
                   limit: 10
+    /sys/dedupe:
+      x-modules:
+        - path: sys/deduplicator.js
+          options:
+            redis: *redis_config
     /sys/purge:
       x-modules:
         - path: sys/purge.js
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -305,7 +305,8 @@ class BaseExecutor {
             if (isRateLimited) {
                 return { status: 200 };
             }
-            return P.each(handler.exec, (tpl) => {
+            return this._dedupeMessage(expander)
+            .then(() => P.each(handler.exec, (tpl) => {
                 const request = tpl.expand(expander);
                 request.headers = Object.assign(request.headers, {
                     'x-request-id': origEvent.meta.request_id,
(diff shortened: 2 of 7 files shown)


gwicke:
A couple of thoughts:

- Separate end point vs. filter: Deduplication (and rate limiting) seem like they might fit the filter interface. The attraction would be that deduplication could be configured quite easily on different end points. I know we discussed this idea before, and you looked into it, so I'm curious what your conclusion was on that direction.
- Mixin vs. composition: In terms of readability and ease of understanding, inheritance and mixins often tend to be worse than composition. In this case, it seems that the main functionality provided is a somewhat hyperswitch-specific redis wrapper. Just making that a plain class member could be easier to understand.
- HTTP interface: The current POST interface returning 200 and true / false seems very RPC-y. A more idiomatic HTTP approach could be a PUT to a key with `If-None-Match: *`. This would also move the key selection to the client.

Pchelolo:
> Separate end point vs. filter: Deduplication (and rate limiting) seem like they might fit the filter interface. The attraction would be that deduplication could be configured quite easily on different end points. I know we discussed this idea before, and you looked into it, so I'm curious what your conclusion was on that direction.

I've been exploring these 2 options for a while now and I couldn't come up with a decent use of filters. The main issue is that a filter must be set on an endpoint - that means we need to introduce one level of indirection on a rule - add a special `rerender_restbase` endpoint for example and then call it from the rule. The problem with this approach is that every rule serenaders RESTBase slightly differently, so creating one single `rerender_restbase` endpoint that fit all the use-cases is VERY tricky. And if we create one per rule, that makes the config completely unreadable, way longer and way trickier to write. I do agree that having a filter would be nice, but I've already spent way too much time trying to fit the filters into the system and they just don't fit in. It's like fitting a cube into a round hole. I think eventually we will come up with a solution, but that could be made as a refactoring step - the core functionality will stay unchanged.

> Mixin vs. composition: In terms of readability and ease of understanding, inheritance and mixins often tend to be worse than composition. In this case, it seems that the main functionality provided is a somewhat hyperswitch-specific redis wrapper. Just making that a plain class member could be easier to understand.

The mixin thingy was my little experiment on how would people react to this kind of multiple inheritance based composition. We already have a delegate approach here in `KafkaFactory` - an object factory that encapsulates all the setup for a Kafka client. I don't really have strong feelings towards any of the options, so whatever everyone thinks is better I will use.

> HTTP interface: The current POST interface returning 200 and true / false seems very RPC-y. A more idiomatic HTTP approach could be a PUT to a key with If-None-Match: *. This would also move the key selection to the client.

Ye, it is kinda an RPC here.. And the idea with a header seems nice, lemme redo that.

Pchelolo:
Corresponding deploy repo config change:  https://gerrit.wikimedia.org/r/374629

PR #199, opened 2017-08-29
Don't attempt to deduplicate reties

Pchelolo:
We shouldn't even try to dedupe retries.

cc @wikimedia/services


gwicke:
:+1: , lets merge once travis is done.

PR #200, opened 2017-08-29
Actually drop duplicated messages

Pchelolo:
Actually begin dropping duplicate messages

cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -2,7 +2,6 @@
 
 const BaseExecutor = require('./base_executor');
 const URI = require('hyperswitch').URI;
-const utils = require('./utils');
 
 /**
  * A rule executor managing matching and execution of a single rule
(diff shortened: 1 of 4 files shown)


lib/rule_executor.js
```diff
@@ -47,16 +46,8 @@ class RuleExecutor extends BaseExecutor {
...
-                this.log('warn/dedupe', {
-                    message: 'Event was deduplicated',
-                    event_str: utils.stringify(expander.message)
-                });
-            }
-        })
-        .catch({ status: 404 }, () => {})
-        .thenReturn(expander);
+        .get('body')
+        .catch({ status: 404 }, () => false);
```
gwicke: Hmmm.. just `false` is not really a valid HTTP response. Perhaps make this `{ body: false }` instead?

Pchelolo: But it's just a function call, we don't call it via http interface

gwicke: Ah, right.. this is the caller, not the callee. Nevermind.

gwicke:
LGTM, lets just wait for Mr. Travis.

Pchelolo:
Ok travis done, merging,

PR #204, opened 2017-09-12
Make the Redis key prefix configurable in a deduplicator

Pchelolo:
Now that we have 2 separate installations of ChangeProp it would be nice to separate the reds keys as well with.a configurable prefix.

cc @wikimedia/services

### DIFF
--- a/sys/deduplicator.js
+++ b/sys/deduplicator.js
@@ -10,9 +10,10 @@ class Deduplicator extends mixins.mix(Object).with(mixins.Redis) {
     constructor(options) {
         super(options);
 
-        this._options = options;
+        this._options = options || {};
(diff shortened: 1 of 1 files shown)


sys/deduplicator.js
```diff
@@ -10,9 +10,10 @@ class Deduplicator extends mixins.mix(Object).with(mixins.Redis) {
     constructor(options) {
         super(options);
 
-        this._options = options;
+        this._options = options || {};
         this._log = this._options.log || (() => {});
         this._expire_timeout = options.window || 86400;
+        this._prefix = this._options.redis_prefix || 'CP';
```
d00rman: A better default would have been the service name as declared in the config file rather than a hardcoded one.

PR #208, opened 2017-09-20
Integrate budgeteer for deduplication & rate limiting

gwicke:
- Create a set of services shared between subscription / rule processors
once in the kafka module, and pass it down into the processors.
- Set up an event context object capturing basic event-derived
information for easy use in multiple utility methods related to
deduplication and rate limiting.
- Fully replace the custom deduplicator with budgeteer.
- Only mark events as processed after actually processing the event
successfully.

## TODO

- Properly test deduplication functionality.
- Double check the handling of root jobs in particular.
- Do a clean up pass.


gwicke:
For discussion only for now. Lets address the TODOs before merging.

config.example.wikimedia.yaml
```diff
@@ -28,11 +28,6 @@ spec: &spec
                 # errors per hour
                 - interval: 3600
                   limit: 10
-    /sys/dedupe:
```
Pchelolo: I thought we were going to run them in parallel for now?

config.example.yaml
```diff
@@ -45,6 +45,11 @@ services:
                   metadata_broker_list: 127.0.0.1:9092
                   dc_name: default
                   concurrency: 10
+                  budgeteer:
```
Pchelolo: I don't think this one needs it - we don't really set up anything in this example. Might as well just delete it.

lib/base_executor.js
```diff
@@ -182,6 +196,85 @@ class BaseExecutor {
         });
     }
 
+    /**
+     * Temporary internal wrapper around limiter & budgeteer.
+     */
+    _reportSuccess(expander, evCtx) {
```
Pchelolo: Perhaps rename `evCtx` to something more understandable?

test/feature/static_rules.js
```diff
@@ -56,7 +56,10 @@ describe('Basic rule management', function() {
             '{}'
         ].map((strMsg) => Buffer.from(strMsg)),
             (msg) => producer.produce('test_dc.simple_test_rule', 0, msg))
-        .then(() => common.checkAPIDone(service))
+        .then(() => {
+            console.log('produced');
```
Pchelolo: Console

sys/deduplicator.js
```diff
@@ -1,99 +0,0 @@
-"use strict";
```
Pchelolo: Again, I thought we were going to run them in parallel?

lib/base_executor.js
```diff
@@ -182,6 +196,85 @@ class BaseExecutor {
...
+                this.hyper.log('trace/dedupe', {
+                    message: 'Event was deduplicated based on sha1/id',
+                    event_str: utils.stringify(message),
+                });
+                return true;
+            }
+            if (!message.root_event) {
+                return false;
+            }
+            const rootEventKey = `${this._prefix}_dedupe_${name}`
```
Pchelolo: What's the difference between this and `rootEventBudgetKey`? Why do we have 2 of them?

Pchelolo:
Made a primary pass over the PR, will do one more a bit later. Also tests are failing.

config.example.wikimedia.yaml
gwicke: I don't think that is worth the complexity, given the differences in focus and approach. Comparing to the job queue deduplication rate seems more relevant.

config.example.yaml
gwicke: You mean, delete the example file?

lib/base_executor.js
gwicke: Renamed it to 'eventCtx'.

config.example.yaml
Pchelolo: Ya, it doesn't really make sense to have one

gwicke: Okay, removed in 9680e42.

config.example.yaml
```diff
@@ -1,53 +0,0 @@
-# Number of worker processes to spawn.
```
d00rman: Why is this file being removed?

lib/base_executor.js
```diff
@@ -29,23 +40,26 @@ class BaseExecutor {
     /**
      * Creates a new instance of a rule executor
      * @param {Rule} rule
-     * @param {KafkaFactory} kafkaFactory
-     * @param {Object} hyper
-     * @param {Function} log
+     * @param {Object} services
```
d00rman: Let's document which services we are after here?

lib/base_executor.js
```diff
@@ -263,15 +355,46 @@ class BaseExecutor {
...
-        const startTime = Date.now();
 
         const expander = {
             message: origEvent,
             match: handler.expand(origEvent)
         };
 
+        // Prefer meta.hash, but also support top-level sha1 for MediaWiki
+        // jobs.
+        let eventHash = origEvent.meta.hash || origEvent.sha1;
```
d00rman: `hash` is not part of an event's `meta` object.

sys/kafka.js
```diff
@@ -19,36 +20,49 @@ const RuleSubscriber = require('../lib/rule_subscriber');
...
+            kafka: new KafkaFactory(options),
+            // Rate limiting and deduplication
+            budgeteer: new Budgeteer(options.budgeteer),
+            hyper: null // Filled in during setup
+        };
 
-        this.subscriber = new RuleSubscriber(options, this.kafkaFactory);
+        // Shorthands
+        this.log = this._services.log;
+
```
d00rman: Nit: no need for empty lines here

config.example.yaml
Pchelolo: There's literally no use for this config - there's no third party installs of change-prop

test/feature/static_rules.js
```diff
@@ -287,6 +289,53 @@ describe('Basic rule management', function() {
...
+        return P.try(() => producer.produce('test_dc.simple_test_rule', 0, eventBuffer))
+        .then(() => {
+            common.checkAPIDone(service);
+            nock.cleanAll();
+            setupService();
+        })
+        // Send the exact same event again, and make sure it is not executed
+        // again.
+        .then(() => producer.produce('test_dc.simple_test_rule', 0, eventBuffer))
+        .then(() => common.checkAPIDone(service))
```
Pchelolo: This test will not really verify correctness. Even if the second event is not deduplicated, the api will be done anyway. To verify that something is not happening we generally do add `.times(2)` to the nock and then use `common.checkPendingMocks(1)` after posting both events.

Also if you do it like that you don't need the `setupService` function and the whole test will become much simpler.

test/feature/static_rules.js
```diff
@@ -287,6 +289,53 @@ describe('Basic rule management', function() {
...
+                    'content-type': 'application/json',
+                    'user-agent': 'ChangePropTestSuite'
+                }
+            })
+            .post('/', {
+                'test_field_name': 'test_field_value',
+                'derived_field': 'test',
+                'random_field': rand,
+            })
+            .reply({});
```
Pchelolo: The signature of the reply function is `.reply(http_status, body)`

test/feature/static_rules.js
```diff
@@ -287,6 +289,53 @@ describe('Basic rule management', function() {
         .finally(() => nock.cleanAll());
     });
 
+    it('Should deduplicate identical events based on sha1', () => {
+        const rand = Math.random().toString();
```
Pchelolo: We've got `common.randomString` for better readability of these randoms. Also if we ever integrate event validation it will not pass as `sha1` is alphanumeric and doesn't allow a dot.

test/feature/static_rules.js
```diff
@@ -287,6 +289,53 @@ describe('Basic rule management', function() {
...
+                common.eventWithProperties('simple_test_rule', {
+                    sha1: rand,
+                    random: rand,
+                    message: 'test'
+                })));
+
+        setupService();
+        return P.try(() => producer.produce('test_dc.simple_test_rule', 0, eventBuffer))
+        .then(() => {
+            common.checkAPIDone(service);
```
Pchelolo: This is no bueno. The `checkAPIDone()` method is async, because we have the `maxAttempts` parameter, so sometimes it returns a promise - here the promise will be kinda lost and also the `cleanAll` might be called before the check actually completes. 

On the other hand if the check throws, the `cleanAll` will never be executed and it will affect other tests and `nock` is global.

One comment below I'm proposing a different approach to checking that something is not executed.

test/feature/static_rules.js
```diff
@@ -56,7 +56,9 @@ describe('Basic rule management', function() {
             '{}'
         ].map((strMsg) => Buffer.from(strMsg)),
             (msg) => producer.produce('test_dc.simple_test_rule', 0, msg))
-        .then(() => common.checkAPIDone(service))
+        .then(() => {
+            return common.checkAPIDone(service);
+        })
```
Pchelolo: What's the point of this change?

lib/base_executor.js
gwicke: Not yet. This is something we can add, though.

sys/kafka.js
gwicke: Done.

test/feature/static_rules.js
gwicke: Just a left-over from some debugging. I changed it back to the short form.

test/feature/static_rules.js
gwicke: Okay, I followed the example of the preceding test here. Updated both.

test/feature/static_rules.js
gwicke: Yeah, good catch on checkAPIDone().

test/feature/static_rules.js
gwicke: This test was working (and the API not done), since deduplication suppressed the second execution. Returning isDone=true would have caused the test to fail.

In any case, the checkPendingMocks method is cleaner overall, so I changed it.

lib/base_executor.js
Pchelolo: @gwicke Why not move the `sha1` property under meta then and start generating it for the EventBus events? Why adding a separate property?

gwicke: I don't think we will necessarily use sha1 for the hash, so I kept the name generic. We could definitely assign the MW job sha1 value to "hash", and move it out of the event content.

PR #213, opened 2017-11-20
Try to speed up ChangeProp

Pchelolo:
Various perf-related fixes:
1. Some minor optimizations of some of the hot spots obtained from the profiler.
2. Try to create a new config option and fetch messages from node-rdkafka with batches. For low-traffic topics batches are not great since it will delay significantly processing of individual message, but for the high traffic topics they might save a lot of CPU on going back and forth to C++ world.

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -24,6 +24,15 @@ const DEFAULT_CONCURRENCY = 30;
  */
 const DEFAULT_COMMIT_INTERVAL = 500;
 
+/**
+ * The default number of messages to consume at once
+ * For low-volume topics this must be one to avoid big delays in consumption.
+ *
+ * @const
+ * @type {number}
+ */
+const DEFAULT_CONSUMER_BATCH_SIZE = 1;
+
 class BaseExecutor {
 
     /**
@@ -46,9 +55,13 @@ class BaseExecutor {
         this.log = log;
         this.options = options;
         this.concurrency = rule.spec.concurrency || this.options.concurrency || DEFAULT_CONCURRENCY;
+        this.consumerBatchSize = rule.spec.consumer_batch_size
+            || this.options.consumer_batch_size
+            || DEFAULT_CONSUMER_BATCH_SIZE;
 
         this._commitTimeout = null;
-        this._pendingMsgs = [];
+        // In order ti filter out the pending messages faster make them offset->msg map
+        this._pendingMsgs = new Map();
         this._toCommit = undefined;
         this._consuming = false;
     }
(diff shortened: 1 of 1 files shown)


lib/base_executor.js
```diff
@@ -80,34 +93,35 @@ class BaseExecutor {
...
+
+                const handler = this.getHandler(message);
+                if (handler) {
+                    // We're pushing it to pending messages only if it matched so that items
+                    // that don't match don't count against the concurrency limit.
+                    this._pendingMsgs.set(msg.offset, msg);
+                    // Note: we don't return the promise here since we wanna process messages
+                    // asynchronously from consuming them to be able to fill up the pendingMsg
+                    // queue and achieve the level of concurrency we want.
+                    this.processMessage(message, handler)
```
d00rman: This now changes the meaning of concurrency: now it means that `concurrency + consumerBatchSize - 1` messages can be processed at once (and will be for high-traffic topics). The concurrency limit should be checked before processing the message.

lib/base_executor.js
```diff
@@ -80,34 +93,35 @@ class BaseExecutor {
...
+                    .finally(() => {
+                        this._notifyFinished(msg);
+                        if (this._pendingMsgs.size < this.concurrency && !this._consuming) {
+                            this._consume();
```
d00rman: Following the above comment, this method should be reworked so that it's possible to process messages without consuming anything.

lib/base_executor.js
Pchelolo: @d00rman I don't think this is an issue, at least definitely not right now. The `consumerBatchSize` will be > 1 only for very high-traffic topics with concurrency > 100 and the idea of the `consumerBatchSize` is to be << `concurrency`. I'm right now thinking to set it to 10 for `RecordLintJob`, so it will create only 10% difference for actual concurrency. Reworking the consumption pipeline in order to achieve the previous concurrency semantics is not easy (and I actually wanna rework it even more significantly in the future to decouple message fetching from message processing a bit more). 

So, before getting into a bigger refactoring I wanna find out if that actually makes a difference in production - locally it does improve things, but the results are not very convincing.

So my plan was to merge this, deploy without increase in consumerBatchSize, see what happens and how other small optimizations affect things, increase consumerBatchSize to 10, see what happens and if consumerBatchSize idea is indeed good - rework the consumption pipeline, if it's not good - revert consumerBatchSize. Perhaps it's actually better to separate the consumerBatchSize from other improvements into 2 separate PRs.

PR #214, opened 2017-11-20
Optimize commited offset choosing

Pchelolo:
Selecting the minimum offset to commit across pending messages is a pretty heavy operation - we need to iterate over the whole pending queue which could be quite big and select the minimal offset. However, we only commit once per commit interval (500ms) - so why don't we select the offset to commit once per 500 ms as well and not update it after each processed message? This can create a bit of a lag in the committed offsets for topics with low traffic, but it's totally fine - we will not reexecute already processed events because now we have redis-based deduplication.

According to my profiling this makes us significantly faster (at least it's one of the major contributors to object allocation and CPU usage in the JavaScript land

Additionally - use lazy-evaluation for trace logging - that was a big contributor to mem/cpu usage as well.

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -62,7 +62,6 @@ class BaseExecutor {
         this._commitTimeout = null;
         // In order ti filter out the pending messages faster make them offset->msg map
         this._pendingMsgs = new Map();
-        this._toCommit = undefined;
         this._consuming = false;
     }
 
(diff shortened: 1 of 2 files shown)


lib/base_executor.js
```diff
@@ -211,33 +210,29 @@ class BaseExecutor {
...
-                if (this._toCommit) {
-                    const committing = this._toCommit;
-                    return this.consumer.commitMessageAsync(committing)
-                    .then(() => {
-                        if (this._toCommit && this._toCommit.offset === committing.offset) {
-                            // Don't commit what we've just committed
-                            this._toCommit = undefined;
+                const minPendingOffset = () => {
+                    const offsetIterator = this._pendingMsgs.keys();
+                    let lowestOffset = Number.MAX_VALUE;
```
d00rman: Given that execution time varies, it looks to me as if it would be possible for `finishedMgs` to actually have the lowest offset, so perhaps we should set `lowestOffest = pendingMsg.offset` here?

d00rman: Ah no, disregard the comment. That cannot work.

PR #222, opened 2017-12-20
Copy the correct config files into the container

Pchelolo:
Obvious mistake.

cc @wikimedia/services


d00rman:
Hehe, I first spotted it, but then forgot about it when I went to look for the mediawiki-containers side of things :)

PR #223, opened 2017-12-20
Rerender mobile-sectionss and summary on MW purge and null edit.

Pchelolo:
Sometimes due to software bugs MCS output could be wrong.
If it's cached, the editors will naturally try to purge it with
MW purge or a null edit. But since HTML might not have changed,
the optimization in RESTBase that doesn't issue purges if HTML's
unchanged will not be issued. This config change fixes the situation.

Bug: T182953

cc @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -164,16 +164,29 @@ spec: &spec
                 limiters:
                   blacklist: 'html:{message.meta.uri}'
                 exec:
-                  method: get
-                  # This even comes directly from MediaWiki, so title is encoded in MW-specific way.
-                  # Re-encode the title in standard `encodeURIComponent` encoding.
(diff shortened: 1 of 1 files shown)


d00rman:
Hm, hm. This will do the correct thing in cases  such as the one listed on the task, but it will create double requests for summary and mobile-sections in every other instance, so I'm a bit wary of doing such a change. We should either have RESTBase emit change events regardless of whether the HTML changed or not or have it never emit them and use this PR. Having both may produce unwanted results and flood the system with requests.

Pchelolo:
@d00rman the rate of `mw-purge` is close to 0, the rate of null edits is about 2/s - so we're creating 4/s more requests here. Not a big deal in my opinion

d00rman:
Oh right. Good point.

Pchelolo:
I'll make a deploy repo counterpart shortly

PR #224, opened 2018-01-17
librdkafka: Use the main APT pool

d00rman:
Bug: [T185016](https://phabricator.wikimedia.org/T185016)
Bug: [T176126](https://phabricator.wikimedia.org/T176126)


Pchelolo:
Superceeded by #225 as we've seen node-rdkafka v1 doesn't really work with librdkafka 0.11+ that is in the main pool this PR alone shouldn't be merged.

PR #225, opened 2018-01-17
Update node-rdkafka driver to version 2

Pchelolo:
Node-rdkafka version 2+ requires librdkafka 0.11+ that we are going
to upgrade to. The new librdkafka supports Kafka 0.9, however, it
first issues a request to Kafka for supported API version but Kafka 0.9
doesn't support this feature yet, so we have to switch it off until we
migrate Kafka to 0.10+

All the unit tests pass locally, but I propose to proceed with caution -
we can install librdkafka 0.11 on the beta cluster and let it run there
for a while. Also, we can randomly kill brokers in beta and see how it reacts.
Local testing of such things is normally non-conclusive.

cc @wikimedia/services @lavagetto

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -16,6 +16,8 @@ const CONSUMER_DEFAULTS = {
     // We don't want the driver to commit automatically the offset of just read message,
     // we will handle offsets manually.
     'enable.auto.commit': 'false',
+    'api.version.request': 'false',
+    'broker.version.fallback': '0.9.0'
 };
 
 const CONSUMER_TOPIC_DEFAULTS = {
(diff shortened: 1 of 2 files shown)


Pchelolo:
I've made a normal set of tests locally - killing a broker, breaking connection to the broker on the network level, a bunch of stress tests with super-high concurrency and locally all looks good, mem usage seem a little bit better then it was with v1, but as always, local results are not very assuring.

Pchelolo:
And travis is dead today ☠️

Pchelolo:
Discussed this with @d00rman so self-merging and deploying to beta.

PR #228, opened 2018-01-19
Rerender summary after mobile-sections.

Pchelolo:
MCS will use cached mobile-sections-lead to generate the summary,
so summary must only be rerendered after the mobile-sections-lead
update is completed.

Bug: T184753
cc @wikimedia/services @wikimedia/mobile

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -81,9 +81,10 @@ spec: &spec
                 cases: # Non wiktionary domains - rerender summary
                   - match:
                       meta:
-                        uri: '/^(?<proto>https?):\/\/[^\/]+\/api\/rest_v1\/page\/html\/(?<title>[^/]+)$/'
+                        uri: '/^(?<proto>https?):\/\/[^\/]+\/api\/rest_v1\/page\/mobile-sections\/(?<title>[^/]+)$/'
                       tags:
                         - restbase
+                        - mobile-sections
                     match_not:
                       - meta:
                           domain: '/wiktionary.org$/'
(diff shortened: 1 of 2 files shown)


config.example.wikimedia.yaml
```diff
@@ -81,9 +81,10 @@ spec: &spec
                 cases: # Non wiktionary domains - rerender summary
                   - match:
                       meta:
-                        uri: '/^(?<proto>https?):\/\/[^\/]+\/api\/rest_v1\/page\/html\/(?<title>[^/]+)$/'
+                        uri: '/^(?<proto>https?):\/\/[^\/]+\/api\/rest_v1\/page\/mobile-sections\/(?<title>[^/]+)$/'
                       tags:
                         - restbase
+                        - mobile-sections
```
berndsi: Just curious. What do the tags do?

berndsi: What do the tags do?

Pchelolo:
DO NOT MERGE! This approach might be wrong, see discussion on https://phabricator.wikimedia.org/T184753#3913843

config.example.wikimedia.yaml
Pchelolo: @berndsi basically they are just for documentation purposes to be able to quickly see where the message is coming from. Also, it's nice matching on them with CP and various other tools.

Pchelolo:
Config change https://gerrit.wikimedia.org/r/#/c/406058/

Pchelolo:
Travis just doesn't want to work today at all. Tests pass locally though

PR #233, opened 2018-02-06
Support delayed job execution.

Pchelolo:
If the delayed job functionality is not configurable (basically we rely on the event having a delay_until field in a very specific format, then delayed job execution becomes trivial the idea is that we dequeue a message, if the message is ready to be executed - we execute. If not - we wait for 10 seconds and put the message back to the queue to give ourselves a chance to process some more messages. 

I'm not quite sure how to create a unit test for this though, but manual testing shows it works great. Also perhaps we need to add some instrumentation and custom metrics to the delayed functionality. And it's interesting whether you think relying on the `delay_until` field format and not making this configurable is ok.

Bug: T186261
cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -307,7 +307,7 @@ class BaseExecutor {
                 }
                 // The exec time is to soon to re-enqueue - just wait and execute
                 return P.delay(timeLeft + 1)
-                .then(() => _exec(origEvent, handler, statDelayStartTime, retryEvent));
+                .then(() => this._exec(origEvent, handler, statDelayStartTime, retryEvent));
             }
         }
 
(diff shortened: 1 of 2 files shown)


Pchelolo:
Another question here is that putting the message back to the same topic will not play well with the metrics that we use - we will basically report the same message over and over again and we will not be able to see the actual job enqueue rate. We could avoid that by using a special new `.delay` topic and subscribing to it by default, but technically that's not required and also will tie up a precious thread from the libuv thread pool.

Pchelolo:
Another question is whether we want to try deduplication upon intermediate dequeing - there's no need to put the event back into the queue if we know that it will get deduplicated on execution - however, this requires some custom deduplication logic just for the delayed jobs and it's not very easy to make - I propose to treat this kind of deduplication as an optimization and implement it later if we find out that trafficfor delayed job topics is a problem.

PR #235, opened 2018-02-23
Provide more information about the reasons why a particular message was retried.

Pchelolo:
cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -380,7 +380,8 @@ class BaseExecutor {
             emitter_id: this._emitterId(),
             retries_left: retriesLeft === undefined ? this.rule.spec.retry_limit : retriesLeft,
             original_event: event,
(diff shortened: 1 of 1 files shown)


lib/base_executor.js
```diff
@@ -380,7 +380,8 @@ class BaseExecutor {
             emitter_id: this._emitterId(),
             retries_left: retriesLeft === undefined ? this.rule.spec.retry_limit : retriesLeft,
             original_event: event,
-            reason: errorRes && errorRes.body && errorRes.body.title
+            reason: errorRes && errorRes.body && (errorRes.body.title || errorRes.body.message),
```
d00rman: Wouldn't it be more useful to prioritise the error's message over its title?

Pchelolo: The error is the instance of HTTPError and there we mostly use title

PR #240, opened 2018-03-15
Purge media endpoints from Varnish when HTML changes.

Pchelolo:
Bug: T189751

cc @wikimedia/services @berndsi


Pchelolo:
Deploy repo counterpart:  https://gerrit.wikimedia.org/r/419789

PR #241, opened 2018-03-20
Support partitioned topic for refreshLinks.

Pchelolo:
Bug: T189738

### DIFF
--- a/sys/partitioner.js
+++ b/sys/partitioner.js
@@ -1,0 +1,59 @@
+"use strict";
+
+class Partitioner {
+    constructor(options) {
+        this._options = options || {};
+        this._log = this._options.log || (() => {});
+
+        if (!options.partition_topic_name) {
+            throw new Error('No partition_topic_name was provided to the partitioner');
+        }
+
+        if (!options.partition_key) {
+            throw new Error('No partition_key was provided to the partitioner');
+        }
+
+        if (!options.partition_map) {
+            throw new Error('No partition_map was provided to the partitioner');
+        }
+    }
+
+    /**
+     * Selects a proper partition and reposts the message to the partitioned topic.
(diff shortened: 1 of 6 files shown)


Pchelolo:
The topic with 8 partitions should be created before deploying this.

d00rman:
This PR seems to contain 63236fcafa01525f3013051c672e7a6f23e8f84f as well, which is weird. Needs a rebase perhaps?

sys/partitioner.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Selects a proper partition and reposts the message to the partitioned topic.
+     * @param {HyperSwitch} hyper
+     * @param {Object} req
+     */
+    repostToPartition(hyper, req) {
+        const event = req.body;
+        const partitionKeyValue = event[this._options.partition_key];
```
d00rman: Minor: this will not work if the partitioning key is not a top-level field, e.g. `meta.domain` wouldn't work here.

sys/partitioner.js
```diff
@@ -0,0 +1,59 @@
...
+        let partition = this._options.partition_map[partitionKeyValue];
+        if (partition === undefined) {
+            partition = this._options.partition_default;
```
d00rman: `options.partition_default` is not checked for existence in the constructor, so we may end up in a situation where the partition is still `undefined`. I would propose to check for it in the constructor and set it to `0` if it's not set.

sys/partitioner.js
Pchelolo: Ye, but we dont need support for that now, so we can add it when we need it

sys/partitioner.js
Pchelolo: Done, the check was added

PR #242, opened 2018-03-20
Purge page metadata endpoint on html change and deletes.

Pchelolo:
cc @wikimedia/services @mdholloway @berndsi


Pchelolo:
Deploy repo counterpart https://gerrit.wikimedia.org/r/420797

PR #243, opened 2018-04-03
Support multi-topic rules.

Pchelolo:
As we've decided, for the JobQueue we need to support multi-topic rules that will share a single worker and this patch does just that.

Changes: 
1. Added support for array-based rules. All the rules that subscribe to multiple topics share a single worker now. This needs adjusting in normal change-prop config, because there we use regex-based rules differently.
2. This changes the metric names - all the topics subscribed to a single rule will share the metric name, so we need to decide what to go with metrics. Whatever we decide we will probably loose metric continuity.
3. This lacks the `exclusive rule` feature that would allow to ignore all the topics that other rules were subscribed to already, but it's pretty easy to implement with a config stanza.

TODO: Add a test for resubscription when the existing topics dynamically change

Bug: https://phabricator.wikimedia.org/T191238
cc @wikimedia/services

### DIFF
--- a/lib/rule_subscriber.js
+++ b/lib/rule_subscriber.js
@@ -53,7 +53,7 @@ class RegexTopicSubscription {
         this._hyper = hyper;
         this._ruleName = ruleName;
         this._ruleSpec = ruleSpec;
-        this._topicTester = ruleSpec.topics || (ruleSpec.topic && [ ruleSpec.topic ])
+        this._topicTester = (ruleSpec.topics || (ruleSpec.topic && [ ruleSpec.topic ]))
         .map((topic) => {
             if (/^\/.+\/$/.test(topic)) {
                 // Ok, we've got a regex topic! Compile the regex.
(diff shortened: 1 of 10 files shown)


lib/rule_subscriber.js
```diff
@@ -42,34 +44,66 @@ class BasicSubscription {
...
-    constructor(options, kafkaFactory, hyper, rule, metadataWatch) {
+    constructor(options, kafkaFactory, hyper, ruleName, ruleSpec, metadataWatch) {
         this._kafkaFactory = kafkaFactory;
         this._options = options;
         this._log = this._options.log || (() => {});
         this._hyper = hyper;
-        this._rule = rule;
+        this._ruleName = ruleName;
+        this._ruleSpec = ruleSpec;
+        this._topicTester = ruleSpec.topics || (ruleSpec.topic && [ ruleSpec.topic ])
```
d00rman: Seems to be missing parenthesis; we want to apply the map regardless of whether `topic` or `topics` is specified, but in its current form, the `.map` is applied only to `ruleSpec.topic`

PR #245, opened 2018-04-04
Fixed the new metrics names. Again

Pchelolo:
cc @wikimedia/services


d00rman:
Damn, how could have be both have missed it? :)

PR #246, opened 2018-04-05
Fixed deduplication key calculation for multi-topic rules

Pchelolo:
cc @wikimedia/services

### DIFF
--- a/lib/rule_executor.js
+++ b/lib/rule_executor.js
@@ -43,8 +43,14 @@ class RuleExecutor extends BaseExecutor {
     }
 
     _dedupeMessage(expander) {
+        let dedupeKey;
+        if (this.rule.topics.length === 1) {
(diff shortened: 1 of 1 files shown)


lib/rule_executor.js
```diff
@@ -43,8 +43,14 @@ class RuleExecutor extends BaseExecutor {
     }
 
     _dedupeMessage(expander) {
+        let dedupeKey;
+        if (this.rule.topics.length === 1) {
```
d00rman: If there is only one element, and that one is a regex, it should not use only the rule name.

Pchelolo: A regex can never be here - if the spec contains a regex it will be subscribed by a regex subscriber that will unfold the regex into a list of topics

d00rman: Ah, right!

PR #248, opened 2018-04-05
Added tests for delayed execution

Pchelolo:
Apparently, we don't need to create the topics before hand.

cc @wikimedia/services


Pchelolo:
@d00rman done

Pchelolo:
> All but the first Travis build fail, not sure why?

Apparently, the idea to not create the topics before tests was not that good

PR #249, opened 2018-04-10
Added support for specifying a sampler per rule and globally

Pchelolo:
We've only supported samper per execution-case basis, but not per-rule (for rules with different cases) and not globally. This adds support for those possibilities.

cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -178,8 +178,8 @@ class Rule {
             };
             if (option.sample) {
                 result.sampler = new Sampler(option.sample);
-            } else if (this.spec.sampler) {
-                result.sampler = new Sampler(this.spec.sampler);
+            } else if (this.spec.sample) {
+                result.sample = new Sampler(this.spec.sample);
             }
             return result;
         });
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -176,14 +176,10 @@ class Rule {
                 expand: matcher.expand,
                 match_not: this._processMatchNot(option.match_not)
             };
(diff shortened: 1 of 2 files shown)


lib/rule.js
```diff
@@ -176,14 +176,10 @@ class Rule {
...
             };
-            // TODO: Support more global sampler. For example, if case of
-            // `cases` stanza is used, we need to be able to fallback to
-            // the more global sample stanza from the higher-leve spec.
-            // TODO: possibly support specifying the `sample` options on the
-            // sys/kafka module level and assign it to all the rules when
-            // they're created in the _subscribeRules method.
             if (option.sample) {
                 result.sampler = new Sampler(option.sample);
+            } else if (this.spec.sampler) {
```
d00rman: Here and below: did you mean(TM) `this.spec.sample` ?

eevans: > Here and below: did you mean(TM) this.spec.sample ?

Yeah, this.

Also, maybe we should include an example in an apropos sample config so we don't forget how to use it later. :)

lib/rule_subscriber.js
```diff
@@ -11,8 +11,8 @@ class BasicSubscription {
         this._kafkaFactory = kafkaFactory;
         this._options = options;
         this._log = this._options.log || (() => {});
+        ruleSpec.sampler =  ruleSpec.sampler || options.sampler;
```
d00rman: `s/sampler/sample`

PR #250, opened 2018-04-11
Split retry and error topics for change-prop and job-queue.

Pchelolo:
Now that we support multi-topic rules we can painlessly transfer without losing any retry messages.

However, when we deploy the multi-topic rules we will lose all the backlog anyway, so I'm not sure if it worths trying to save some retries on this one.

Also, the message validation tests were incorrect - they passed even if they should not have. Now they are correct, but they fail because the message.meta.uri doesn't conform to uri pattern. I'm wondering what to do with that - imaging some other uri or just remove the uri pattern from the schemas? cc @ottomata 

Bug: T181291
cc @wikimedia/services


d00rman:
> Now that we support multi-topic rules we can painlessly transfer without losing any retry messages.
> 
> However, when we deploy the multi-topic rules we will lose all the backlog anyway, so I'm not sure if it worths trying to save some retries on this one.

+1, I think we can just drop the old ones. Alternatively, we can deploy, wait a day and then remove them, but given the low rate of retries I wonder if it's worth the effort.

> Also, the message validation tests were incorrect - they passed even if they should not have. Now they are correct, but they fail because the message.meta.uri doesn't conform to uri pattern. I'm wondering what to do with that - imaging some other uri or just remove the uri pattern from the schemas?

Just changing the URI in the tests seems the most straightforward thing to do.

Pchelolo:
> +1, I think we can just drop the old ones. Alternatively, we can deploy, wait a day and then remove them, but given the low rate of retries I wonder if it's worth the effort.

I forgot that the cpjobqueue already is migrated to the multi-topic rules, so I think we can deploy it as is just because I've already coded up the fallback

You mean just rewriting the uri in the test to something like `https://wikipedia.org`? Seems like cheating..

d00rman:
> You mean just rewriting the uri in the test to something like https://wikipedia.org? Seems like cheating..

Why? Rety topics contain the full URI, don't they?

d00rman:
> Not always, if we can't parse the JSON (and that's what we're testing for in tests) the error topic will have /error in the URI :(

So we correct that too, then? IMHO, having the `format: uri` in the schema is helpful, even though I do concede that the validator seems to be testing for URLs, not URIs (so the real bug is there if you ask me).

Pchelolo:
So that is the question - what URL(I) to put there if the message is unparsable and we don't know the original one?

d00rman:
> So that is the question - what URL(I) to put there if the message is unparsable and we don't know the original one?

Something generic like `https://mediawiki.org/wiki/ChangePropagation/error` ?

PR #251, opened 2018-04-12
Record HTML errors from MW in retry message

Pchelolo:
Example when this happened: T192085

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -409,7 +409,7 @@ class BaseExecutor {
     }
 
     _constructRetryMessage(event, errorRes, retriesLeft, retryEvent) {
-        return {
+        const result = {
             meta: {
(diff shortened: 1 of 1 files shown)


lib/base_executor.js
```diff
@@ -420,9 +420,15 @@ class BaseExecutor {
             emitter_id: this._emitterId(),
             retries_left: retriesLeft === undefined ? this.rule.spec.retry_limit : retriesLeft,
             original_event: event,
-            reason: errorRes && errorRes.body && (errorRes.body.title || errorRes.body.message),
             error_status: errorRes && errorRes.status
         };
+        if (errorRes && errorRes.body && typeof res.body === 'object') {
```
d00rman: Ooops, `s/res/errorRes/`

PR #255, opened 2018-04-18
Update dependencies

Pchelolo:
Not ready to update `mocha` - it now needs the tests to shut doesn't cleanly and for change-prop clean shutdown requires some more work.

Not ready to update AJV - it now supports JSON-schema draft-v07 and apparently our draft-04 schemas are not entirely valid.

node-rdkafka should be fixed, the newer version breaks the build on mac.

cc @wikimedia/services


Pchelolo:
Test pass, self merging since it's mostly dev deps update

PR #256, opened 2018-04-25
Support `exclude_topics` config stanza for array and regex rules

Pchelolo:
cc @wikimedia/services


d00rman:
I may be missing something obvious here, but how will this work for regex rules?

PR #262, opened 2018-05-02
Improve logging and adapt to the new hyper logger exposure.

Pchelolo:
1. Stop using hyper.log and use the logger instead.
2. Include the rule name in all the logs
3. Include the topic name in the log whenever possible
4. Fix up some logging bugs.
5. Stop including the rule name in the log level, instead, make the log level descriptive on what's happened and include the rule name as a separate property.

Depends on the hyperswitch 0.10 changes
cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "bluebird": "^3.5.1",
     "cassandra-uuid": "^0.0.2",
     "extend": "^3.0.1",
-    "hyperswitch": "^0.9.6",
+    "hyperswitch": "^0.10.0",
     "service-runner": "^2.6.1",
     "json-stable-stringify": "^1.0.1",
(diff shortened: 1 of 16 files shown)


d00rman:
Needs a rebase now. Figures :)

Pchelolo:
Done with a rebase

package.json
```diff
@@ -38,7 +38,7 @@
     "bluebird": "^3.5.1",
     "cassandra-uuid": "^0.0.2",
     "extend": "^3.0.1",
-    "hyperswitch": "^0.9.6",
+    "hyperswitch": "^0.10.0",
```
d00rman: Since 0.10.1 is already out, let's require it.

Pchelolo: Done

PR #265, opened 2018-05-07
Allow protocol version negotiation.

Pchelolo:
As a part of upgrade of the kafka brokers from 0.9 to 1.1.0
we need to allow the clients to use a new feature of API
version negotiation so that change-prop actually use the
right protocol talking to the brokers.

Also, update CI configuration to test agains 1.1.0 broker version.

Bug: T167039

cc @ottomata @wikimedia/services

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - g++-4.8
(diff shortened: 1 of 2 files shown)


Pchelolo:
Only deploy in the correct moment according to the deployment plan in https://phabricator.wikimedia.org/T167039

PR #266, opened 2018-05-15
Emit a revision-score event

Pchelolo:
Emit a revision-score event after doing the precache request to ORES.

cc @wikimedia/services @ottomata


sys/ores_updates.js
```diff
@@ -0,0 +1,73 @@
...
+                scores: []
+            };
+            const domainScores = res.body[newMessage.database];
+            const revScores = domainScores.scores[`${newMessage.rev_id}`];
+            Object.keys(revScores).forEach((modelName) => {
+                newMessage.scores.push({
+                    model_name: modelName,
+                    model_version: domainScores.models[modelName].version,
+                    prediction: revScores[modelName].prediction,
+                    probability: revScores[modelName].probability
```
ottomata: I wonder if we also need to add any other unschemaed parts of the response.  Aaron made us use additionalProperties: true here, since the score could be anything.  I'm pretty sure he was saying that the scores fields themselves wouldn't have variable keys.

On the other hand, if we could avoid pushing arbitrary data here, I'd much prefer it.  Perhaps all scores we currently use only have prediction and/or probability, and that is good enough?

sys/ores_updates.js
```diff
@@ -0,0 +1,76 @@
...
+                scores: []
+            };
+            const domainScores = res.body[newMessage.database];
+            const revScores = domainScores.scores[`${newMessage.rev_id}`];
+            Object.keys(revScores).forEach((modelName) => {
+                newMessage.scores.push({
+                    model_name: modelName,
+                    model_version: domainScores.models[modelName].version,
+                    prediction: revScores[modelName].score.prediction,
+                    probability: revScores[modelName].score.probability
```
ottomata: Halfak says 'not all scores have a probability'.  We should probably be safe here and only set prediction/probability if the .score has those fields.

d00rman:
Why send the event to EventBus instead of Kafka directly?

Pchelolo:
> Why send the event to EventBus instead of Kafka directly?

Because these events will be exposed publicly via EventStreams and we want to validate them against the schema.

Pchelolo:
In prod we actually update ores both in eqiad and codfw, so made the updater module take a list of uris

Pchelolo:
ok, it's been too long. I'll merge and deploy in beta

PR #273, opened 2018-05-31
Purge definitions endpoint on mediawiki purges.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T192157
cc @wikimedia/services @berndsi

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -182,37 +182,71 @@ spec: &spec
               # RESTBase update jobs
               mw_purge:
                 topic: resource_change
-                match:
-                  meta:
-                    uri: '/^(?<proto>https?):\/\/[^\/]+\/wiki\/(?<title>.+)$/'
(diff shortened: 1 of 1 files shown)


d00rman:
Hm, so if the content has changed, then we will end up with two `resource_change` events - one from RB and one from change prop, won't we?

While that may not be as problematic since we have deduplication, it is semantically wrong to emit the event that the RB resource has changed when in fact it hasn't. Note that the same thing could be achieved by RB emitting the event in either case, which illustrates the problem in doing so.

I don't have a magic solution for this, but we obviously need to think how will we approach this.

Pchelolo:
@d00rman This is a hack indeed - here I'm trying to reuse all the other rules in order not to copy-paste  for wiktionary vs non-wiktionary domains. Although, maybe copy-paste would be fine here as that will help to avoid posting an artificial event.

Pchelolo:
Updated.

PR #274, opened 2018-06-06
Rely on the request template to specify the timeout.

Pchelolo:
cc @wikimedia/services

### DIFF
--- a/lib/rule.js
+++ b/lib/rule.js
@@ -323,7 +323,7 @@ class Rule {
             req.headers = req.headers || {};
             req.followRedirect = false;
             req.retries = 0;
-            req.timeout = this.spec.timeout;
+            req.timeout = req.timeout || this.spec.timeout;
 
             if (!this.spec.decode_results) {
                 req.encoding = null;


lib/rule.js
```diff
@@ -323,7 +322,7 @@ class Rule {
             req.headers = req.headers || {};
             req.followRedirect = false;
             req.retries = 0;
-            req.timeout = this.spec.timeout;
+            req.timeout = req.timeout || DEFAULT_REQUEST_TIMEOUT;
```
d00rman: Hm, this breaks the backwards compatibility. What we want here is:

```js
  req.timeout = req.timeout || this.spec.timeout;
```
This, ofc, implies leaving L159 above intact.

Also, I don't understand why for the case we have in mind this is needed at all? The videoscaler jobs have their own rule, so we can specify the timeout there.

lib/rule.js
```diff
@@ -323,7 +322,7 @@ class Rule {
             req.headers = req.headers || {};
             req.followRedirect = false;
             req.retries = 0;
-            req.timeout = this.spec.timeout;
+            req.timeout = req.timeout || this.spec.timeout || DEFAULT_REQUEST_TIMEOUT;
```
d00rman: Heh, I actually think it's better to leave L159 as is and then just do `req.to || this.spec.to` here. My reasoning is that all the other defaults are set up there, so it gives the enforced time out visibility and improves code readability.

Pchelolo: Done

PR #275, opened 2018-06-19
Assume the wikidata description might change on revert.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T197626

сс @wikimedia/services

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -527,7 +527,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/wbeditentity|wbsetdescription|undo/'
+                  comment: '/wbeditentity|wbsetdescription|undo|Reverted/'
                   rev_content_changed: true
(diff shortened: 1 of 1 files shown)


config.example.wikimedia.yaml
```diff
@@ -527,7 +527,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/wbeditentity|wbsetdescription|undo/'
+                  comment: '/wbeditentity|wbsetdescription|undo|Reverted/'
```
berndsi: Not saying it's wrong. Just curious where do you see `Reverted` in the [history](https://www.wikidata.org/w/index.php?title=Q315507&action=history)? I only see `Restore revision` for the [currently latest change](https://www.wikidata.org/w/index.php?title=Q315507&type=revision&diff=698199571&oldid=697709679).

Pchelolo: Ouch....

Pchelolo: https://github.com/wikimedia/change-propagation/pull/276

PR #276, opened 2018-06-19
Set the correct word in the regex

Pchelolo:
Foolowup for #275

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -527,7 +527,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/wbeditentity|wbsetdescription|undo|Reverted/'
+                  comment: '/wbeditentity|wbsetdescription|undo|restore/'
                   rev_content_changed: true
                 exec:
                   method: post


config.example.wikimedia.yaml
```diff
@@ -527,7 +527,7 @@ spec: &spec
                   page_namespace: 0
                   # It's impossible to modify a comment in wikidata while editing the entity.
                   # TODO: This is a temp solution until we get a more general fragment support T148079
-                  comment: '/wbeditentity|wbsetdescription|undo|Reverted/'
+                  comment: '/wbeditentity|wbsetdescription|undo|restore/'
```
berndsi: Shouldn't those be case-insensitive? Have you tested this with a test item, let's say on test.wikidata.org?

d00rman: The comments in these cases are set by wikidata itself, which uses these exact words.

berndsi: @d00rman Where do you see these strings? I just like to be able to see them as well. I don't see them in https://test.wikidata.org/w/index.php?title=Q112644&action=history.

d00rman: [The revision info](https://test.wikidata.org/w/api.php?action=query&prop=revisions&revids=396312&format=json&formatversion=2) MW API gives you the comment:

```json
{
  "batchcomplete": true,
  "query": {
    "pages": [
      {
        "pageid": 168592,
        "ns": 0,
        "title": "Q112644",
        "revisions": [
          {
            "revid": 396312,
            "parentid": 396311,
            "minor": false,
            "user": "StBernard8",
            "timestamp": "2018-06-20T15:39:06Z",
            "comment": "/* restore:0||396308|StBernard8 */ my own comment during restore"
          }
        ]
      }
    ]
  }
}
```

d00rman: To clarify, the comment in this ^ form comes to CP via the event emitted by MW.

berndsi: Got it. Thanks! That makes sense. 👍

PR #280, opened 2018-07-13
Fix up change-prop blacklisting

Pchelolo:
Needs a followup in the config to remove `^` from regexes.

cc @wikimedia/services


d00rman:
Hmmm, what about special regex chars that would get escaped?

Pchelolo:
> Hmmm, what about special regex chars that would get escaped?

Qué?

PR #284, opened 2018-08-17
Create metrics to track the actual concurrent job executions.

Pchelolo:
The number of concurrent jobs is not the same as the rate
of execution. When the jobs become slow, we can actually
pile up more concurrent jobs and so more DB connections
and HHVM threads even if the execution rate goes down.

This metric will allow us to see if that's indeed happening.

Bug: https://phabricator.wikimedia.org/T202107

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -481,6 +481,8 @@ class BaseExecutor {
                 if (messageDeduped) {
                     return { status: 200 };
                 }
+                this._hyper.metrics.increment(
+                    `${this.statName(origEvent.meta.topic)}_concurrency`);
                 return P.each(handler.exec, (tpl, index) => {
                     const request = tpl.expand(expander);
                     request.headers = Object.assign(request.headers, {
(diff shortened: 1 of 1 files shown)


d00rman:
How is this metric different from the `_exec` one on L504?

Also, I see two problems here:
- this metric includes messages that would get discarded and not actually executed
- sending increment and decrement messages so often might create a lot of UDP traffic, perhaps it would be better to have an internal counter and send it value once per interval

Pchelolo:
> How is this metric different from the _exec one on L504?

This one also decrements, so allows us to have not the rate of executions but the number of concurrent executions.

> this metric includes messages that would get discarded and not actually executed

Only for rate limited messages, deduplications and other edge cases, main tests were done on the message already. Also, we want to graph the pending queue size, this is more accurate.

> sending increment and decrement messages so often might create a lot of UDP traffic, perhaps it would be better to have an internal counter and send it value once per interval

I think it's a premature optimization, we batch them under the hood in the library. we can easily add this if we encounter issues.

d00rman:
> Only for rate limited messages, deduplications and other edge cases, main tests were done on the message already. Also, we want to graph the pending queue size, this is more accurate.

This doesn't seem correct to me. This PR adds code the `_consume()` which in turn calls `processMessage()` which then calls `_exec()` and the latter is the place that all these tests are done, so it looks to me like we are counting the messages here, not the number of executions.

> I think it's a premature optimization, we batch them under the hood in the library. we can easily add this if we encounter issues.

Yes, correct. Either way, since we will be using this only in CP4JQ, could we have this under a config flag so as to minimise the amount of traffic sent? Or do you think this could be useful for CP itself too?

Pchelolo:
> This doesn't seem correct to me. This PR adds code the _consume() which in turn calls processMessage() which then calls _exec() and the latter is the place that all these tests are done, so it looks to me like we are counting the messages here, not the number of executions.

Updated to only use it for actually executing messages.

> Yes, correct. Either way, since we will be using this only in CP4JQ, could we have this under a config flag so as to minimise the amount of traffic sent? Or do you think this could be useful for CP itself too?

I think this metric will be useful in general - we'd be able to have more precise limits both for CP and CP4JQ

Pchelolo:
Self-merging per @d00rman +1

PR #285, opened 2018-08-20
Make the queue sized report stats once a second

Pchelolo:
@d00rman was correct in #284 - we need to accumulate the stats locally - it's not only more efficient, just sending increment-decrement into statsd simply doesn't work.

cc @wikimedia/services


Pchelolo:
ok, I'm too impatient, I wanna see those stats to progress. Will self merge-deploy.

PR #287, opened 2018-08-20
Clone the event in the partitioner.

Pchelolo:
The partitioner actually modifies the original event
which messes up with higher level of execution - metrics,
logging, subsequent requests could be broken. In particular,
our metrics are reporting incorrect data because of this.

This will create some additional cpu/mem usage, but I think
that would be negligible.

cc @wikimedia/services

### DIFF
--- a/sys/partitioner.js
+++ b/sys/partitioner.js
@@ -28,13 +28,19 @@ class Partitioner {
      * @param {Object} req
      */
     repostToPartition(hyper, req) {
-        const event = req.body;
+        let event = req.body;
         const partitionKeyValue = event[this._options.partition_key];
         let partition = this._options.partition_map[partitionKeyValue];
         if (partition === undefined) {
             partition = this._options.partition_default;
         }
 
+        // Clone the event and meta sub-object to avoid modifying
+        // the original event since that could mess up processing
+        // in the executor regarding metrics, limiters, follow-up
+        // executions etc.
+        event = Object.assign({}, event);
+        event.meta = Object.assign({}, event.meta);
         event.meta.topic = this._options.partition_topic_name;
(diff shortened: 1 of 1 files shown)


sys/partitioner.js
```diff
@@ -28,13 +28,19 @@ class Partitioner {
...
         if (partition === undefined) {
             partition = this._options.partition_default;
         }
 
+        // Clone the event and meta sub-object to avoid modifying
+        // the original event since that could mess up processing
+        // in the executor regarding metrics, limiters, follow-up
+        // executions etc.
+        event = Object.assign({}, event);
+        event.meta = Object.assign({}, event.meta);
```
d00rman: This is a work-around for a full clone, but we need a full clone. Let's use `extend` here rather than `Object.assign` since we might need other parts of the event fully cloned too at some point and using `extend` reduces the magic in the code.

Pchelolo: It is a workaround for a full clone. It's a micro-optimization - we know we only change the `meta.topic` here, so we can avoid the full clone.

d00rman: This time is `meta.topic`, but my point is that it's easy to categorise this as "we already clone the object" and commit mistakes later on that are not so easy to spot. Granted, only certain events have other nested objects and we never touch these event-specific properties, but I'm still not convinced.

Let's at the very least put a big fat XXX comment.

d00rman: Also, given that we are constrained by other, external parts that are slowing CP down, we can waste a cycle or two on proper cloning, even though it's an operation that would be done often.

Pchelolo: DONE

PR #288, opened 2018-08-20
Don't clean up the queue size counters

Pchelolo:
Obvious mistake. I was so anxious to get it out that didn't think it through properly.

cc @wikimedia/services


Pchelolo:
Ok, 4th time is a charm. I'll try to deploy this again.

PR #290, opened 2018-08-30
Add rule for sending purge requests on map tile resource_change

mdholloway:
TODO: Get the unit test passing

### DIFF
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -646,6 +646,21 @@ spec: &spec
                     match_not: *others_match_not
                     exec: *page_images_others_rerender
 
+              # Map tile cache invalidation
+              purge_map_tile:
+                topic: 'resource_change'
+                match:
+                  meta:
+                    uri: '/\/osm(-intl)?\/\d{1,2}\/\d{1,6}\/\d{1,6}.png$/'
+                  tags:
+                    - tilerator
+                exec:
+                  method: post
+                  uri: '/sys/purge/'
+                  body:
+                    - meta:
+                        uri: '{{match.meta.uri}}'
+
 num_workers: 0
 logging:
   name: changeprop
--- a/config.example.wikimedia.yaml
+++ b/config.example.wikimedia.yaml
@@ -650,8 +650,6 @@ spec: &spec
               purge_map_tile:
                 topic: 'resource_change'
                 match:
-                  meta:
-                    uri: '/\/osm(-intl)?\/\d{1,2}\/\d{1,6}\/\d{1,6}.png$/'
                   tags:
                     - tilerator
                 exec:
@@ -659,7 +657,7 @@ spec: &spec
                   uri: '/sys/purge/'
                   body:
                     - meta:
-                        uri: '{{match.meta.uri}}'
+                        uri: '{{message.meta.uri}}'
 
 num_workers: 0
 logging:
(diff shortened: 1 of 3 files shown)


config.example.wikimedia.yaml
```diff
@@ -646,6 +646,21 @@ spec: &spec
                     match_not: *others_match_not
                     exec: *page_images_others_rerender
 
+              # Map tile cache invalidation
+              purge_map_tile:
+                topic: 'resource_change'
+                match:
+                  meta:
+                    uri: '/\/osm(-intl)?\/\d{1,2}\/\d{1,6}\/\d{1,6}.png$/'
```
Pchelolo: Will all the tilerator URIs match this pattern or it's a subset of all the possible URIs? If it's the former - no need for a match.

config.example.wikimedia.yaml
```diff
@@ -646,6 +646,21 @@ spec: &spec
...
+                  tags:
+                    - tilerator
+                exec:
+                  method: post
+                  uri: '/sys/purge/'
+                  body:
+                    - meta:
+                        uri: '{{match.meta.uri}}'
```
Pchelolo: {{message.meta.uri}} is better. However, if the URI in the `resource_change` will not be protocol-relative, we need to update purge.js line 38 to just replace https with http in the protocol in the uri

test/feature/update_rules.js
```diff
@@ -1072,5 +1072,48 @@ describe('RESTBase update rules', function() {
...
+                    closed = true;
+                    done();
+                }
+            } catch (e) {
+                udpServer.close();
+                closed = true;
+                done(e);
+            }
+        });
+        udpServer.bind(4321);
```
Pchelolo: Since we now have multiple of these in several separate tests, could you please wrap it into some function and call it from both?

Pchelolo:
I believe if you follow my comments the test should pass. Especially regarding `{{message.meta.uri}}`

config.example.wikimedia.yaml
mdholloway: All tilerator URLs should match, so just using the tag should be fine.

config.example.wikimedia.yaml
mdholloway: Might as well keep it consistent and send with the protocol.  Updating.

mdholloway: Wait, nm, I misread that and it looks like it *should* be protocol-relative, which is what I was in fact sending to begin with.

Pchelolo: yup. either protocol-relative or we'd need to change the. `purge.js` module a bit.

Other events include the protocol in the uri, so it might be better to be consistent

test/feature/update_rules.js
mdholloway: Done

mdholloway: Done

Pchelolo:
Fabulous.

Does it already produce events? Also needs a corresponding config change for the deploy repo.

mdholloway:
I just merged https://github.com/kartotherian/tilerator/pull/41 which provides for sending events.  (Unfortunately we're blocked on deploying to production until the maps cluster is updated to Stretch, so this won't actually happen there for a little while, but we can deploy to beta.)  I'll get the deploy config changes going now.

Pchelolo:
I guess you'd also need puppet changes to provide the uri of the eventbus service to tilerator.

mdholloway:
Yep!

Pchelolo:
ok. So we can deploy the whole thing to beta and test, nothing stops us from deploying this to prod as well, so that everything will work automatically when tilerator is deployed

PR #291, opened 2018-09-10
Log the failed event execution with warn level.

Pchelolo:
Bug: [T203929](https://phabricator.wikimedia.org/T203929)

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -584,6 +584,15 @@ class BaseExecutor {
                 topic: message.meta.topic
             }));
             return reportError();
+        } else {
+            this._logger.log('warn/exec_error', () => ({
+                message: `Exec error in ${this._hyper.config.service_name}`,
+                status: e.status,
+                stack: e.stack,
+                event_str: utils.stringify(message),
+                topic: message.meta.topic,
+                error_body: e.body
+            }));
         }
 
         if (!this.rule.shouldIgnoreError(e)) {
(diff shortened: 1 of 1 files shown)


d00rman:
I think this will make logging too aggressive as it will report all errors - both original ones and retries, so we'll have multiple log entries per message. Logging it only for the original message might be a solution, but I still worry that that would be too much.

Aren't we logging retry failures already?

Pchelolo:
With retries not being reported, as Tim suggested, it might be pretty hard to debug. This is warn level, so will only exist on the node in the log file and not reported to logstash, so I think it's a manageable amount of logs. The retry rate is not very high, should be fine.

We only log retry failures when we exceed the maximum number of retry attempts.

d00rman:
> With retries not being reported, as Tim suggested, it might be pretty hard to debug. This is warn level, so will only exist on the node in the log file and not reported to logstash, so I think it's a manageable amount of logs. The retry rate is not very high, should be fine.

In that case, `s/warn/info/` is needed, warn is reported to logstash, info is not.

> We only log retry failures when we exceed the maximum number of retry attempts.

Right, but this PR changes that and logs all `HTTPError` occurrences since it is executed as part of the `_catch()` method, which is called whenever a non-2xx status is received. This is my worry and why I think we should be careful here.

Speaking of being careful, why do we need the stack trace? If it's a `HTTPError`, then we know we received it as a response, in which case the stack is superfluous, IMHO.

Pchelolo:
Ok, nade it `info` and made it only log for errors we do not explicitly ignore. 

I think the volume of the logs should be manageable.

PR #293, opened 2018-10-03
Properly resubscribe when the list of topics has changed.

Pchelolo:
Tested and tested and tested in labs and it works!

Bug: T199444

cc @wikimedia/services

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -106,6 +106,7 @@ class BaseExecutor {
         this._pendingMsgs = new Set();
         this._pendingCommits = new Map();
         this._consuming = false;
+        this._connected = false;
     }
 
     subscribe() {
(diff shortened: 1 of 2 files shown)


lib/base_executor.js
```diff
@@ -134,6 +136,10 @@ class BaseExecutor {
     }
 
     _consume() {
+        if (!this._connected) {
+            return;
```
d00rman: Should the msg maps and _consuming flag be cleared here as well?

Pchelolo: Re msg maps - even after re-subscribing we still want to commit what we needed to commit to topics. 

Re _consuming - if we're disconnecting because of exiting - we don't care. If we're resubscribing, this loop will be broken, the next loop will be started and we will reset _consuming to 'true' on the next line

Pchelolo: Eventually, I wanna abstract away a `LimitedConcurrencyConsumer` and simplify this code

PR #297, opened 2018-12-03
Move the errors out of the scores

Pchelolo:
cc @ottomata


Pchelolo:
I have run the tests locally and Travis is not being nice today, so merging.

PR #298, opened 2018-12-04
Don't provide scores or errors if there aren't any elements

ottomata:
Bug: T210465

### DIFF
--- a/sys/ores_updates.js
+++ b/sys/ores_updates.js
@@ -71,9 +71,7 @@ class OresProcessor {
                 };
                 if (revScores[modelName].error) {
                     Object.assign(score, revScores[modelName].error);
-                    if (newMessage.errors === undefined) {
-                        newMessage.errors = [];
-                    }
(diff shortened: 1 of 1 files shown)


sys/ores_updates.js
```diff
@@ -69,6 +71,9 @@ class OresProcessor {
                 };
                 if (revScores[modelName].error) {
                     Object.assign(score, revScores[modelName].error);
+                    if (newMessage.errors === undefined) {
```
Pchelolo: newMessage.errors = newMessage.errors || []

sys/ores_updates.js
```diff
@@ -83,6 +88,9 @@ class OresProcessor {
                         name: probName,
                         value: originalScore.probability[probName]
                     }));
+                    if (newMessage.scores === undefined) {
```
Pchelolo: Idem

Pchelolo:
Nit picking as always

PR #299, opened 2018-12-06
Added performer to the revision-scores event

Pchelolo:
We have forgotten to add a performer property to the revision-score event.

cc @wikimedia/services @ottomata


Pchelolo:
This is trivial. Self-merge

PR #303, opened 2019-01-18
Removal of mocha-eslint and eslint-config-node-services deps.

hknustwmf:
Minor formatting updates. Added .eslintrc for style exceptions


package.json
```diff
@@ -51,18 +52,15 @@
...
     "js-yaml": "^3.11.0",
     "nock": "^9.2.5",
     "preq": "^0.5.6",
     "ajv": "^5.5.2",
-    "mocha-eslint": "^4.1.0",
-    "eslint": "^4.19.1",
-    "eslint-config-node-services": "^2.2.5",
-    "eslint-config-wikimedia": "^0.5.0",
-    "eslint-plugin-jsdoc": "^3.6.3",
+    "eslint": "^5.12.0",
```
Pchelolo: I believe the eslint-config-wikimedia already depends on eslint, no need to explicitly add this dep here.

package.json
```diff
@@ -51,18 +52,15 @@
...
+    "eslint-config-wikimedia": "^0.10.0",
+    "eslint-plugin-jsdoc": "^4.0.0",
```
Pchelolo: I think we might want to upstream this dep to the config.

hknustwmf: plugin-jsdoc? Is that something I need to do? Or hold off on that for now?

Pchelolo: Nothing to do now, just a mental note for when we will be upstreaming some of the overrides

PR #307, opened 2019-03-07
Add blubber pipeline stanzas

akosiaris:
Add .pipeline/blubber.yaml

Bug: T213193
Change-Id: I80e1ba75ab8054e75cca48609fb140faeb0c1f7d

### DIFF
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -1,0 +1,27 @@
+version: v3
+base: docker-registry.wikimedia.org/nodejs-slim
+apt: { packages: [librdkafka++1, librdkafka1] }
+lives:
+  in: /srv/service
+runs:
+  environment: { APP_BASE_PATH: /srv/service }
+
+variants:
+  build:
+    base: docker-registry.wikimedia.org/nodejs-devel
+    apt: { packages: [librdkafka-dev] }
+    node: { requirements: [package.json]}
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: 0 } }
+  development:
+    includes: [build]
+    entrypoint: [node, server.js]
+  test:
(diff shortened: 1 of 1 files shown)


akosiaris:
Travis complains because:
```
E: Package 'oracle-java8-set-default' has no installation candidate
```

not sure how to fix it.

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,27 @@
...
+apt: { packages: [librdkafka++1, librdkafka1] }
+lives:
+  in: /srv/service
+runs:
+  environment: { APP_BASE_PATH: /srv/service }
+
+variants:
+  build:
+    base: docker-registry.wikimedia.org/nodejs-devel
+    apt: { packages: [librdkafka-dev] }
```
d00rman: For the build step, we need this package to come from our WMF repo.

.pipeline/blubber.yaml
```diff
@@ -0,0 +1,27 @@
...
+    node: { requirements: [package.json]}
+    runs: { environment: { LINK: g++ } }
```
d00rman: Also needs the `BUILD_LIBRDKAFKA=0` env var set.

d00rman:
cf. [package.json's build stanza](https://github.com/wikimedia/change-propagation/blob/bb78865d392ebdbcfde623fdbc13358bfe0dccaf/package.json#L69-L88) for the build reqs.

.pipeline/blubber.yaml
akosiaris: That happens anyway in the pipeline. In the images, `apt.wikimedia.org` is always preferred, same way as in production

akosiaris:
The repo comment did not need addressing, I did address the ENV var one.

PR #309, opened 2019-03-15
Create a mock kafka factory to enable unit tests without Kafka

Pchelolo:
cc @wikimedia/services


lib/kafka_factory.js
```diff
@@ -291,4 +291,17 @@ class KafkaFactory {
         return this._createProducerOfClass(GuaranteedProducer, logger);
     }
 }
-module.exports = KafkaFactory;
+
+let GLOBAL_FACTORY;
```
d00rman: Would it not be better to have the factory to use as a param to `create` instead of global vars?

.travis.yml
```diff
@@ -21,8 +21,7 @@ addons:
 
 before_script:
 - npm run install-kafka
-- npm run start-kafka
 
-script: npm run coverage && (npm run coveralls || exit 0)
+script: npm test && npm run start-kafka && npm run coverage && (npm run coveralls || exit 0)
```
d00rman: `npm run coverage` will run `npm test` again, won't it?

Pchelolo: coverage runs 'mocha' with no MOCK_KAFKA env variable - thus it's an integration test.

lib/kafka_factory.js
Pchelolo: Param to create? We need to store the global factory somewhere if we want to replace it from tests preparation code

PR #312, opened 2019-03-21
Enable librdkafka error logging for a producer.

Pchelolo:
cc @ottomata

### DIFF
--- a/lib/kafka_factory.js
+++ b/lib/kafka_factory.js
@@ -28,7 +28,7 @@ class GuaranteedProducer extends kafka.Producer {
     /**
      * @inheritdoc
      */
-    constructor(conf, topicConf) {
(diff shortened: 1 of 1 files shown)


lib/kafka_factory.js
```diff
@@ -42,6 +42,10 @@ class GuaranteedProducer extends kafka.Producer {
         this.on('ready', () => {
             this._pollInterval = setInterval(() => this.poll(), 10);
         });
+
+        this.on('event.error', e => logger.log('error/producer', {
+            msg: 'Producer error', e
```
d00rman: IIRC bunyan has a special error formatter that works on `err` keys, so this should be `err: e`

Pchelolo:
Done

PR #313, opened 2019-03-26
Support templates in partitioned topics names

Pchelolo:
Bug https://phabricator.wikimedia.org/T219159

### DIFF
--- a/sys/partitioner.js
+++ b/sys/partitioner.js
@@ -1,13 +1,14 @@
 'use strict';
 
 const extend = require('extend');
+const Template = require('hyperswitch').Template;
 
 class Partitioner {
     constructor(options) {
         this._options = options || {};
 
-        if (!options.partition_topic_name) {
-            throw new Error('No partition_topic_name was provided to the partitioner');
+        if (!options.templates || !options.templates.partition_topic) {
+            throw new Error('No partition_topic template was provided to the partitioner');
         }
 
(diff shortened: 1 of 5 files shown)


sys/partitioner.js
```diff
@@ -21,6 +22,8 @@ class Partitioner {
         if (!options.partition_default) {
             throw new Error('No partition_default was provided to the partitioner');
         }
+
+        this._partitionedTopicTemplate = new Template(options.templates.partition_topic);
```
d00rman: There is no alternative provided here if `partition_topic_name` is provided directly (without a template), which means `options.templates` will be `undefined`, resulting in a run-time error.

Pchelolo: We drop support for partition_topic_name and it's checked in the beginning of the constructor.

d00rman: Ah ok, indeed I misread the `if`-block. In this case, let's please bump the minor version of CP.

Pchelolo: Done

d00rman: Heh, I was thinking of bumping the minor version, not the patch one, but ok. We shall do that too, one day, one day :)

d00rman:
Published as [v0.9.2](https://github.com/wikimedia/change-propagation/tree/v0.9.2)

PR #316, opened 2019-05-31
T223427 - Replace ChangeProp service with TesttServer from service-runner

hknustwmf:
I linted the tests and replaced the service.


.eslintrc.yml
```diff
@@ -1,6 +1,10 @@
 extends: 
   - 'eslint-config-wikimedia/server'
 
+env:
+  mocha: 
+    - true
+
```
Pchelolo: Nonono... Let's add `.eslitrc.yml` under test, exend this one and allow `mocha` only under tests. Just like https://github.com/wikimedia/restbase/blob/master/test/.eslintrc.yml

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "change-propagation",
-  "version": "0.9.2",
+  "version": "0.9.3",
```
Pchelolo: We're not really changing any production code, no need for a version bump

test/feature/job_rules.js
```diff
@@ -7,25 +7,26 @@ const P = require('bluebird');
...
 
     const changeProp = new ChangeProp('config.jobqueue.wikimedia.yaml');
     let producer;
 
-    before(function() {
+    before(function () {
         // Setting up might take some tome, so disable the timeout
         this.timeout(50000);
         return changeProp.start()
+        // eslint-disable-next-line
```
Pchelolo: Given that we will have special eslintrc for tests we can allow console.log in tests and remove this

test/feature/job_rules.js
```diff
@@ -37,8 +38,10 @@ describe('JobQueue rules', function() {
             .post('/wiki/Special:RunSingleJob', sampleEvent)
             .reply({});
             return producer.produce(`test_dc.mediawiki.job.${jobType}`, 0, sampleEvent.toBuffer())
-            .then(() => common.checkAPIDone(service))
-            .finally(() => nock.cleanAll());
+            .then(() => common.checkAPIDone(service, 10))
```
Pchelolo: Why 10?

test/feature/job_rules.js
```diff
@@ -37,8 +38,10 @@ describe('JobQueue rules', function() {
...
+            .finally(() => {
+                nock.cleanAll();
+            });
```
Pchelolo: What's the point in changing this to contain a block `{}`?

test/feature/job_rules.js
```diff
@@ -54,8 +57,8 @@ describe('JobQueue rules', function() {
         })
         .post('/wiki/Special:RunSingleJob', sampleEventCopy)
         .reply({});
-        return producer.produce(`test_dc.mediawiki.job.refreshLinks`, 0, sampleEvent.toBuffer())
-        .then(() => common.checkAPIDone(service))
+        return producer.produce('test_dc.mediawiki.job.refreshLinks', 0, sampleEvent.toBuffer())
+        .then(() => common.checkAPIDone(service, 10))
```
Pchelolo: Why are we changing i to 10?

test/feature/static_rules.js
```diff
@@ -9,19 +9,20 @@ const P = require('bluebird');
...
     const changeProp = new ChangeProp('config.test.yaml');
 
     let producer;
 
-    before(function() {
+    before(function () {
         // Setting up might take some tome, so disable the timeout
         this.timeout(30000);
         return changeProp.start()
+        /* eslint-disable-next-line no-console */
```
Pchelolo: Idem about eslint rules override.

test/feature/update_rules.js
```diff
@@ -35,18 +34,19 @@ describe('update rules', function() {
                 }
             });
         })
-        .then((res) => siteInfoResponse = res.body)
+        .then((res) => { siteInfoResponse = res.body; })
+        /* eslint-disable-next-line no-console */
```
Pchelolo: Idem about special eslint rule

test/utils/changeProp.js
```diff
@@ -1,52 +1,36 @@
...
+    }
+
+    start() {
+        const delay = process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_START_DELAY;
 
-var ChangeProp = function(configPath) {
-    this._running = false;
-    this._configPath = configPath;
-};
+        setTimeout(() => super.start(), delay);
```
Pchelolo: This is not equivalent to what it used to be. Before we were starting the service first, then delaying for CHANGE_PROP_START_DELAY before returning a promise, so that the tests would be run after a delay that CP needs to connect/negotiatiate with Kafka. Now you're doing it vice-versa.

test/utils/changeProp.js
```diff
@@ -1,52 +1,36 @@
...
-            return P.delay(1000).then(() => this.start());
+
+    stop() {
+        if (this._running) {
+            return this._runner.stop()
+            .tap(() => {
+                common.clearKafkaFactory();
+                this._running = false;
+            })
+            .delay(process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_STOP_DELAY);
```
Pchelolo: Why are you not reusing `super.stop` here?

test/utils/common.js
```diff
@@ -8,7 +8,7 @@ const assert       = require('assert');
 const preq         = require('preq');
 const yaml         = require('js-yaml');
 const Ajv          = require('ajv');
-const mock_require = require('mock-require');
+const mockRequire = require('mock-require');
```
Pchelolo: The `=` signs were aligned, not anymore.

test/utils/common.js
```diff
@@ -181,6 +182,7 @@ common.events = {
         dt = new Date().toISOString()
     ) {
         const domain = /https?:\/\/([^/]+).+/.exec(uri)[1];
+        // eslint-disable-next-line no-useless-escape
```
Pchelolo: Why are you disabling useless-escape lint rule instead of removing a useless escape?

test/feature/job_rules.js
hknustwmf: Arbitrary number. The default is set to 50. 50 x 0.5 sec is 25 seconds but the test timeout is set to 10 seconds. Changed it to 20 so it matches the test timeout.

test/feature/job_rules.js
hknustwmf: I probably had additional code in there that I later removed. Removed the block

test/feature/job_rules.js
hknustwmf: see above

test/feature/job_rules.js
Pchelolo: Given that we have 10 seconds timeout everywhere, let's just make the default to 20. 50 was an arbitrary number as well :)

.eslintrc.yml
hknustwmf: Split up now

package.json
hknustwmf: Reversed it to 0.9.2

test/feature/job_rules.js
hknustwmf: Removed

test/feature/job_rules.js
hknustwmf: Changed it to 19 so the retries fail prior to timeout. Otherwise you wouldn't know what went wrong other that the test timed out

test/utils/changeProp.js
hknustwmf: Returning a promise from start() with a delayed resolve

test/utils/changeProp.js
hknustwmf: Calling super.stop() instead now.

test/utils/common.js
hknustwmf: It was incorrect for one regex. I removed the 2 disable line where the linter was correct.

test/utils/changeProp.js
```diff
@@ -1,52 +1,34 @@
...
-};
+    start() {
+        const delayInMs = process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_START_DELAY;
 
-ChangeProp.prototype.start = function() {
-    if (this._running) {
-        console.log('The test server is already running. Skipping start.')
-        return P.resolve();
+        super.start();
+        return new P.Promise((resolve) => setTimeout(resolve, delayInMs));
```
Pchelolo: I meant something like
```
start() {
 return super.start().delay(process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_START_DELAY);
}
```

test/utils/changeProp.js
```diff
@@ -1,52 +1,34 @@
...
-            common.clearKafkaFactory();
-            this._running = false;
-        })
-        .delay(process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_STOP_DELAY);
+    stop() {
+        if (this._running) {
+            return super.stop()
+            .tap(() => {
+                common.clearKafkaFactory();
+            })
```
Pchelolo: Again.. you know I hate unnesessary block syntax for pone-liners.

test/utils/changeProp.js
```diff
@@ -1,52 +1,34 @@
...
-ChangeProp.prototype.stop = function() {
-    if (this._running) {
-        return this._runner.stop()
-        .tap(() => {
-            common.clearKafkaFactory();
-            this._running = false;
-        })
-        .delay(process.env.MOCK_SERVICES ? 0 : CHANGE_PROP_STOP_DELAY);
+    stop() {
+        if (this._running) {
```
Pchelolo: Meh.. this is not really needed. If the test code is faulty and calls 'stop' multiple times, clearing the factory multiple times would be fine. And not relying on internal `_running` property would be kinda nice. 

We probably should reject to double-start/double-stop the service in service-runner eventually.

Pchelolo:
Please rebase on #317 so that we have Travis on our team for this one.

Pchelolo:
Hm... The tests pass locally... Please rebase so that we could see what's up with travis.

Pchelolo:
The test failures are definitely unrelated, I'll look into the test separately. LGTM!

Please resolve the ticket

PR #317, opened 2019-06-03
Use openjdk in travis instead of oracle jdk

hknustwmf:
The oracle-java8-set-default package is no longer available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - g++-4.8
-      - oracle-java8-set-default
+      - openjdk-8-jre-headless
 
 before_script:
 - npm run install-kafka


.travis.yml
```diff
@@ -17,7 +17,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
       - g++-4.8
-      - oracle-java8-set-default
+      - oracle-java9-set-default
```
Pchelolo: We are running Java 8 in production, so I'd prefer that we still do tests with kafka running the same Java major version. Let's do what we do for RESTBase and install `openjdk-8-jre-headless`

Pchelolo:
-1

.travis.yml
hknustwmf: Changed to OpenJDK 8

Pchelolo:
Now, let's wait for Travis to confirm we are right here.

Pchelolo:
Gosh... I'll retire before it finishes

Pchelolo:
ok... Travis is failing, but it's definitely not related to this particular PR. I'll merge this one and we can figure out the failures in the next one. This at least gives us actual test execution in Travis back.

PR #319, opened 2019-06-24
Prepare for switch to eventgate and support .meta.stream

Pchelolo:
Bug: https://phabricator.wikimedia.org/T226198

### DIFF
--- a/config.test.yaml
+++ b/config.test.yaml
@@ -83,15 +83,15 @@ spec: &spec
                   uri: '/sys/queue/events'
                   body:
                     - meta:
-                        topic: '{{message.produce_to_topic}}'
+                        stream: '{{message.produce_to_topic}}'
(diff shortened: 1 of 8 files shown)


Pchelolo:
The tests are failing for unrelated reason. Investigating now.

d00rman:
From what I can tell, all of the test events now use `stream`. It would be good to have a mix of the two given that we'll be in that situation for a while.

Pchelolo:
> From what I can tell, all of the test events now use `stream`. It would be good to have a mix of the two given that we'll be in that situation for a while.

Ok, makes sense. Reverted in some places, so now it's a fair mix.

Pchelolo:
ok, finally I got travis to pass. We probably need to merge and deploy soon to unblock @ottomata

PR #320, opened 2019-06-24
Fix mocks for job queue tests

Pchelolo:
- After fixing how templates expand the URI in swagger-router, we needed to fix the mocks
- The last test is about delays, so wait a bit longer before quitting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ addons:
       - openjdk-8-jre-headless
 
 before_script:
+- export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/
 - npm run install-kafka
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -20,6 +20,7 @@ addons:
       - openjdk-8-jre-headless
 
 before_script:
+- java -version
```
d00rman: Probably not needed?

Pchelolo: I'm debugging the build. it fails cause travis now ships java 11 and we actually don't switch to java 8 and it all fails because the kafka version we run doesn't support java 11.

Pchelolo:
Ok, finally...

PR #326, opened 2019-08-07
Remove workaround for ORES not supporting eventgate events.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T228688


ottomata:
+1 fwiw :)

PR #327, opened 2019-08-07
Support both meta.stream and meta.topic in the partitioner.

Pchelolo:
Bug: https://phabricator.wikimedia.org/T230048

### DIFF
--- a/test/utils/common.js
+++ b/test/utils/common.js
@@ -218,14 +218,14 @@ common.jobs = {
     get updateBetaFeaturesUserCounts() {
         return {
             __proto__: eventMethods,
+            $schema: 'mediawiki/job/1.0.0',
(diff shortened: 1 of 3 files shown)


test/utils/common.js
```diff
@@ -305,6 +306,7 @@ common.jobs = {
         const releaseTimestamp = Date.now() / 1000 + 3;
         return {
             __proto__: eventMethods,
+            $schema: 'mediawiki/job/1.0.0',
             database: 'commonswiki',
             delay_until: `${releaseTimestamp}`,
```
ottomata: delay_until is a date-time string in new event

Pchelolo: Yup. And that's why I'm gonna make one more pull request

PR #330, opened 2019-09-18
T225211: Modify OresProcessor to conform to new revision/score schema

clarakosi:


Pchelolo:
Tests are failing!

Pchelolo:
> @Pchelolo is there a way we can trigger this code in dev to test it with the new schema? In mediawiki-vagrant perhaps?

No. It's downloads the schema from girhub :)

clarakosi:
The tests are passing for Node 10 but failing for Node 6 because Node 6 doesn't support async/await as shown [here](https://node.green/#ES2017-features-async-functions). I guess we could drop support for Node 6 or go back and change the functions to use promises or even use Babel.

package.json
```diff
@@ -17,7 +17,8 @@
     "integration": "export UV_THREADPOOL_SIZE=128 && npm run lint && npm run cleanup && mocha",
     "lint": "eslint --cache --max-warnings 0 --ext .js .",
     "coverage": "export UV_THREADPOOL_SIZE=128 && npm run cleanup && nyc --reporter=lcov mocha",
-    "coveralls": "cat ./coverage/lcov.info | coveralls"
+    "coveralls": "cat ./coverage/lcov.info | coveralls",
+    "postinstall": "$(npm bin)/jsonschema-tools install-git-hook"
```
Pchelolo: No need for the git hook for sure.

Pchelolo:
Seems good, let's see what Travis say

PR #331, opened 2019-09-30
Take job domain into account when deduplicating.

Pchelolo:
For the job queue the message sha1 is describing the job itself,
but might not include the domain which the job is intended for.
I don't think we'd ever want to deduplicate jobs cross-wiki,
so include message domain into a deduplication key.

Bug: https://phabricator.wikimedia.org/T234226

### DIFF
--- a/sys/deduplicator.js
+++ b/sys/deduplicator.js
@@ -24,11 +24,12 @@ class Deduplicator extends mixins.mix(Object).with(mixins.Redis) {
     checkDuplicate(hyper, req) {
         const name = req.params.name;
         const message = req.body;
+        const prefix = `${this._prefix}_dedupe_${name}`;
 
         // First, look at the individual event duplication based on ID
         // This happens when we restart ChangeProp and reread some of the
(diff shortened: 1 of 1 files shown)


sys/deduplicator.js
```diff
@@ -99,7 +100,8 @@ class Deduplicator extends mixins.mix(Object).with(mixins.Redis) {
                 return sha1Duplicate;
             }
 
-            const rootEventKey = `${this._prefix}_dedupe_${name}_${message.root_event.signature}`;
+            const rootEventKey =
+                `${prefix}_${name}_${message.meta.domain}_${message.root_event.signature}`;
```
d00rman: Won't this cause the root event key to be unique per wiki rather than overall? Like, we have a root job that spawns another job for another wiki, then `message.meta.domain` will point to the _target_ wiki, not the _source_ wiki, won't it?

Pchelolo: oh. indeed.

PR #332, opened 2019-11-26
Support dotted JSONPath notation for partitioner key

Pchelolo:
We need that for partitioning cirrusSearchElasticWrite job since it's key is inside an object.

### DIFF
--- a/sys/partitioner.js
+++ b/sys/partitioner.js
@@ -27,6 +27,18 @@ class Partitioner {
     }
 
     /**
+     * Returns a partition key value based on a configured partition key
+     * @param {Object} event
+     * @return {undefined|string}
+     * @private
+     */
+    _getPartitionKeyValue(event) {
+        const result = this._options.partition_key.split('.')
+        .reduce((value, key) => value ? value[key] : undefined, event);
+        return typeof result === 'string' ? result : undefined;
+    }
+
+    /**
      * Selects a proper partition and reposts the message to the partitioned topic.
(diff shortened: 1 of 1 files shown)


sys/partitioner.js
```diff
@@ -26,6 +26,18 @@ class Partitioner {
...
 
+    /**
+     * Returns a partition key value based on a configured partition key
+     * @param {Object} event
+     * @return {undefined|string}
+     * @private
+     */
+    _getPartitionKeyValue(event) {
+        const value = this._options.partition_key.split('.')
+        .reduce((value, key) => value ? value[key] : undefined, event);
```
d00rman: For clarity, it would be better to pick a different accumulator name (or change the name of the variable on the line above).

sys/partitioner.js
```diff
@@ -34,8 +46,7 @@ class Partitioner {
      */
     repostToPartition(hyper, req) {
         const origEvent = req.body;
-        const partitionKeyValue = origEvent[this._options.partition_key];
-        let partition = this._options.partition_map[partitionKeyValue];
+        let partition = this._options.partition_map[this._getPartitionKeyValue(req.body)];
```
d00rman: For clarity, `s/req.body/origEvent/`

sys/partitioner.js
Pchelolo: done

sys/partitioner.js
Pchelolo: done

PR #333, opened 2019-12-04
K8s changes

hknustwmf:


package.json
```diff
@@ -70,16 +70,16 @@
     "node": ">=6"
   },
   "deploy": {
-    "node": "6.11.1",
+    "node": "6.17.1",
```
d00rman: AFAIK k8s uses 6.11.1 packaged by us

package.json
```diff
@@ -70,16 +70,16 @@
...
     "target": "debian",
     "env": {
-      "BUILD_LIBRDKAFKA": "0"
+      "BUILD_LIBRDKAFKA": "1"
```
d00rman: librdkafka should not be built.

.pipeline/blubber.yaml
```diff
@@ -8,10 +8,12 @@ runs:
 
 variants:
   build:
+    apt: { packages: [librdkafka-dev, git, build-essential, python-dev] }
     base: docker-registry.wikimedia.org/nodejs-devel
-    apt: { packages: [librdkafka-dev] }
+    copies: [local]
```
d00rman: The `local` variant does not exist in this file...

package.json
hknustwmf: I checked the version in the image. I can doublecheck

package.json
hknustwmf: No it needs to be built to work. I tried this a few times and the images packages does not work without rebuilding

.pipeline/blubber.yaml
hknustwmf: As far I understand it it just copies the code over. Correct me if I am wrong

package.json
hknustwmf: I stand corrected on that one. The devel version has 6.17 not the the slim prod version.

package.json
```diff
@@ -70,16 +70,16 @@
     "node": ">=6"
   },
   "deploy": {
```
Pchelolo: This whole section AFAIK is not used in k8s, only to build the deploy repo for snap-based deployments, so why are you making changes to it? If it needs to be changed for some reason, it should be done in a separate PR.

package.json
Pchelolo: See comment above

package.json
```diff
@@ -70,16 +70,16 @@
...
     "env": {
-      "BUILD_LIBRDKAFKA": "0"
+      "BUILD_LIBRDKAFKA": "1"
     },
     "dependencies": {
       "debian": [
         {
           "repo_url": "https://apt.wikimedia.org/wikimedia",
-          "release": "jessie-wikimedia",
+          "release": "stretch-wikimedia",
```
Pchelolo: See comment above

.pipeline/blubber.yaml
Pchelolo: Yeah, I think that is the case..

.pipeline/blubber.yaml
```diff
@@ -8,10 +8,12 @@ runs:
...
     node: { requirements: [package.json]}
-    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: 0 } }
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "1" }, insecurely: true }
+    entrypoint: [node, server.js]
```
Pchelolo: I don't think 'build' requires any entry point.

.pipeline/blubber.yaml
```diff
@@ -8,10 +8,12 @@ runs:
 
 variants:
   build:
+    apt: { packages: [librdkafka-dev, git, build-essential, python-dev] }
     base: docker-registry.wikimedia.org/nodejs-devel
-    apt: { packages: [librdkafka-dev] }
+    copies: [local]
     node: { requirements: [package.json]}
-    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: 0 } }
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "1" }, insecurely: true }
```
Pchelolo: Why are we building librdkafka? We have librdkafka-dev, so we don't need to build it. We use Debian package to link to, if we build it from sources we'd be linking to statically linked librdkafka from node-rdkafka, and not using the deb package. If node-rdkafka doesn't build, that means the versions are incompatible. (the librdkafka version in the docker-registry.wikimedia.org/nodejs-devel image is not compatible with node-rdkafka version from package.json). You can confirm that by checking which librdkafka version do you have in the docker image. That would be an interesting problem to solve, but let's first confirm it is the problem.

Pchelolo: What does 'insecurely' mean here and why do we need to be insecure?

.pipeline/blubber.yaml
```diff
@@ -8,10 +8,12 @@ runs:
 
 variants:
   build:
+    apt: { packages: [librdkafka-dev, git, build-essential, python-dev] }
```
Pchelolo: Why do we need git and python-dev?

package.json
hknustwmf: Meaning what? Try it yourself

.pipeline/blubber.yaml
hknustwmf: That I am wrong? I'll do some more research on this

.pipeline/blubber.yaml
hknustwmf: Hmm. Check the blubber output. Let me know

.pipeline/blubber.yaml
hknustwmf: The lib refs in the image don't match the library version that gets pulled. Part of the problem I had to get this put together.

hknustwmf: insecurely (boolean) 
should the application in the container be run as a user that can't write anything to the filesystem, including caches, or as a user that can? Production variants should have this set to `false`, but other variants may set it to `true` in some circumstances.

hknustwmf: When I ran the tests they failed due to cp writing locally

.pipeline/blubber.yaml
hknustwmf: Because it won't build otherwise

.pipeline/blubber.yaml
Pchelolo: No, that you're right. I think that it just copies the code over is the case

package.json
Pchelolo: > This whole section AFAIK is not used in k8s, only to build the deploy repo for snap-based deployments, so why are you making changes to it? If it needs to be changed for some reason, it should be done in a separate PR.

package.json
Pchelolo: > This whole section AFAIK is not used in k8s, only to build the deploy repo for snap-based deployments, so why are you making changes to it? If it needs to be changed for some reason, it should be done in a separate PR.

.pipeline/blubber.yaml
Pchelolo: I mean, it adds `ENTRYPOINT ["node", "server.js"]` but I don't think we're actually expected to use that entrypoint in the build variant, we're basing the other variants on this variant and overriding the entrypoints

.pipeline/blubber.yaml
Pchelolo: Regarding 'insecure' - let's add it to the end variants that need it. If 'test' need it - let's add it to 'test'. If 'build' variant builds without it (and my suspicion it does) - let's not add it. 

Regarding the packages versions mismatch.. That is a problem, but building from sources is not the solution. Our nom distribution model is not amazing, but relying on deb packages where we can is at least better then allowing node-pre-gyp download binaries from amazon :) 

So, the problem is that currently change-prop runs on Jessie, where we have librdkafka 0.11.3, while the nodejs-devel distro depends on stretch, where we have  librdkafka 0.11.6 (see https://apt.wikimedia.org/wikimedia/dists/ ) and I guess it doesn't build with 0.11.6. 

Every version of node-rdkafka is tied to a specific version of librdkafka, for that reason we pin to exact version of node-rdkafka in package.json. We depend on 2.3.1 that depends on 0.11.3, and support for 0.11.6 was added in 2.5.1 [AFAIK](https://github.com/Blizzard/node-rdkafka/commit/56c31c4e81f2a042666160338ad65dc4f8f2d87e#diff-04c6e90faac2675aa89e2176d2eec7d8). So I guess that if you update node-rdkafka to 2.5.1 (maybe 2.5.2, I'm not exactly sure how their release process works) - it should build using the librdkafka from Debian.

Once that works, we will have a problem with a version mismatch in k8s vs current installation, but I guess we can think about it once we get there. My idea would be that the pipeline works from Gerrit anyway, so for the transition period we push your PR with the newer version to gerrit and use pipeline from there, while maintaining the sources for current deployments via scapulas over in GitHub. Or use different branches in hopes that scab-based deployment will die soon enough.

hknustwmf: It builds fine with the librdkafka package but segfaults at runtime. 2.6.1 is the earliest version that supports 0.11.6. The image only contains libssl1.1 if that makes a difference.

.pipeline/blubber.yaml
hknustwmf: You are right. Not needed.

.pipeline/blubber.yaml
hknustwmf: I guess for completeness I should provide more information

service-runner@2.7.3 depends on

 ├─┬ gc-stats@1.4.0
│ │ └─┬ node-pre-gyp@0.13.0

which forces a node-gyp rebuild which fails with 
```
Traceback (most recent call last):
  File "/usr/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 13, in <module>
    import gyp
  File "/usr/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 8, in <module>
    import gyp.input
  File "/usr/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.py", line 5, in <module>
    from compiler.ast import Const
ImportError: No module named compiler.ast
```

if python-dev is not installed

.pipeline/blubber.yaml
Pchelolo: Ok, I figured it. You need node-rdkafka 2.6.1, then you need to state BUILD_LIBRDKAFKA=0, and also you need `libsasl2-dev` package on the build stage. `libsasl2-dev` is the key

hknustwmf: Added the libsasl2-dev package but it still blows up for me

Pchelolo: I have your PR checked out. 

```
ppchelko@Petrs-MBP-2 ~/w/change-propagation (hknust)> git diff
diff --git a/.pipeline/blubber.yaml b/.pipeline/blubber.yaml
index 7820bda..b5f8e82 100644
--- a/.pipeline/blubber.yaml
+++ b/.pipeline/blubber.yaml
@@ -8,11 +8,11 @@ runs:
 
 variants:
   build:
-    apt: { packages: [librdkafka-dev, git, build-essential, python-dev] }
+    apt: { packages: [librdkafka-dev, libsasl2-dev, git, build-essential, python-dev] }
     base: docker-registry.wikimedia.org/nodejs-devel
     copies: [local]
     node: { requirements: [package.json]}
-    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "1" }, insecurely: true }
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "0" }, insecurely: true }
     entrypoint: [node, server.js]
   development:
     includes: [build]
diff --git a/package.json b/package.json
index 4595017..bf52722 100644
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "hyperswitch": "^0.12.4",
     "mediawiki-title": "^0.6.5",
     "murmur-32": "^0.1.0",
-    "node-rdkafka": "2.3.1",
+    "node-rdkafka": "2.6.1",
     "ratelimit.js": "^1.8.0",
     "redis": "^2.8.0",
     "service-runner": "^2.7.1",
```

```
ppchelko@Petrs-MBP-2 ~/w/change-propagation (hknust)> ~/go/bin/blubber .pipeline/blubber.yaml test > Dockerfile; cat Dockerfile 
FROM docker-registry.wikimedia.org/nodejs-devel
USER "root"
ENV HOME="/root"
ENV DEBIAN_FRONTEND="noninteractive"
RUN apt-get update && apt-get install -y "librdkafka++1" "librdkafka1" "librdkafka-dev" "libsasl2-dev" "git" "build-essential" "python-dev" && rm -rf /var/lib/apt/lists/*
RUN groupadd -o -g "65533" -r "somebody" && useradd -l -o -m -d "/home/somebody" -r -g "somebody" -u "65533" "somebody" && mkdir -p "/srv/service" && chown "65533":"65533" "/srv/service" && mkdir -p "/opt/lib" && chown "65533":"65533" "/opt/lib"
RUN groupadd -o -g "900" -r "runuser" && useradd -l -o -m -d "/home/runuser" -r -g "runuser" -u "900" "runuser"
USER "somebody"
ENV HOME="/home/somebody"
WORKDIR "/srv/service"
ENV APP_BASE_PATH="/srv/service" BUILD_LIBRDKAFKA="0" LINK="g++"
COPY --chown=65533:65533 ["package.json", "./"]
RUN npm install
COPY --chown=65533:65533 [".", "."]
ENV NODE_ENV=""
ENTRYPOINT ["npm", "test"]
LABEL blubber.variant="test" blubber.version="+"
```

```
ppchelko@Petrs-MBP-2 ~/w/change-propagation (hknust)> docker build -t cp_test .
<bla bla bla>
```

```
ppchelko@Petrs-MBP-2 ~/w/change-propagation (hknust)> docker run cp_test

> change-propagation@0.9.3 test /srv/service
> export MOCK_SERVICES=true && npm run lint && mocha


> change-propagation@0.9.3 lint /srv/service
> eslint --cache --max-warnings 0 --ext .js .



  JobQueue rules
    ✓ Should propagate updateBetaFeaturesUserCounts job (504ms)
<bla bla bla>
```

Pchelolo: What error does it blow up with?

.dockerignore
```diff
@@ -0,0 +1,10 @@
+# Remove potentially existing node_modules folder from build context
+.git/*
+.gitignore
+.eslintcache
+.mocharc.yaml
```
Pchelolo: For the `test` target this is needed.

.dockerignore
```diff
@@ -0,0 +1,10 @@
...
+.travis.yml
+.vscode/*
+node_modules
+LICENSE
+README.md
```
Pchelolo: There's a bunch of stuff in `.gitignore` that would be nice to add here.

.pipeline/blubber.yaml
```diff
@@ -8,10 +8,11 @@ runs:
 
 variants:
   build:
+    apt: { packages: [librdkafka-dev, build-essential, python-dev, git, libsasl2-dev] }
     base: docker-registry.wikimedia.org/nodejs-devel
-    apt: { packages: [librdkafka-dev] }
+    copies: [local]
     node: { requirements: [package.json]}
-    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: 0 } }
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "0" }, insecurely: false }
```
Pchelolo: Ok, I've pinned down the SEGfault to this 'insecurely' flag. If I turn it into 'true', everything goes correctly.

Pchelolo: Even more interesting: we can build with 'insecurely: false', but to run tests, we'd need 'insecurely:true' and then it does work.

Pchelolo: Ok. Never mind... This only matters because with `insecurely:false` the npm test process can't read the MOCK_KAFKA env variable and actually tries to load up librdkafka/node-rdkafka and segfaults. So, the fact I've seen it working is not really relevant unfortunately.

package.json
hknustwmf: Right. I reversed this.

package.json
hknustwmf: I reversed all changes

.dockerignore
hknustwmf: The CI process checks out from git and these files should not be present but for development convenience I added them to .dockerignore

.pipeline/blubber.yaml
hknustwmf: Added insecure to the test variant and also added --recursive to mocha otherwise no tests could be found.

.dockerignore
hknustwmf: I missed that. Mocha couldn't find the tests which worked before.

package.json
hknustwmf: Reversed

.pipeline/blubber.yaml
hknustwmf: Switched to 10 as discuessed

.pipeline/blubber.yaml
Pchelolo: The `--recursive` for mocha lives in `.mocharc.yaml`. If it's not docker ignored, it should be picked up.

package.json
```diff
@@ -63,14 +63,14 @@
...
-    "preq": "^0.5.7",
+    "preq": "^0.5.14",
     "redis-mock": "^0.43.0"
   },
   "engines": {
     "node": ">=6"
   },
   "deploy": {
-    "node": "6.11.1",
+    "node": "6.11.0",
```
Pchelolo: why?

.pipeline/blubber.yaml
```diff
@@ -8,20 +8,22 @@ runs:
...
 variants:
   build:
-    base: docker-registry.wikimedia.org/nodejs-devel
-    apt: { packages: [librdkafka-dev] }
+    apt: { packages: [librdkafka-dev, build-essential, python-dev, git, libsasl2-dev] }
+    base: docker-registry.wikimedia.org/nodejs10-devel
+    copies: [local]
     node: { requirements: [package.json]}
-    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: 0 } }
+    runs: { environment: { LINK: g++, BUILD_LIBRDKAFKA: "0" }, insecurely: false }
```
Pchelolo: Is the default for 'insecurely' false? if so, we can remove the explicit setting.

.gitignore
```diff
@@ -1,4 +1,3 @@
-/Dockerfile
```
Pchelolo: We still wanna ignore the Dockerfile. Why are you removing it?

package.json
hknustwmf: By mistake

.pipeline/blubber.yaml
hknustwmf: Agreed

.gitignore
hknustwmf: Reversed

Pchelolo:
This looks good to me!

Lemme play with it for a little, and unless something strange happens, this seems to be ready for a merge

Pchelolo:
> Lemme play with it for a little, and unless something strange happens, this seems to be ready for a merge

Hm... It works and it is ready to be merged and it all looks good but! this changes the version of node-rdkafka, so if we merge it into master, we will no longer be able to build a proper deploy repo and deploy on SCB!

Here's my proposal.
1. We change this PR to point to `k8s` branch on GitHub.
2. We leave master on GH intact, and use it for building the deploy repo
3. We push the `k8s` branch from GH to the master branch on gerrit since that's used by pipeline to build the containers. 

This is not great, but it will work. It would be nice to do the switch as quick as possible so that we don't live in the messy inverse for long. 
Could you please change the target branch of this PR to `k8s` branch (or whatever name you find appropriate) and then I'll merge it and push it into gerrit master.

hknustwmf:
I don't see a k8s branch on GH. Do you want me to branch master and name it k8s?

Pchelolo:
> I don't see a k8s branch on GH. Do you want me to branch master and name it k8s?

Yeah. Create a new `k8s` branch out of current master, push it in wikimedia/change-propagation and change the target of this PR from `wikimedia/master` to `wikimedia/k8s`

PR #336, opened 2020-02-07
Create GH actions manifest

Pchelolo:


Pchelolo:
Ok, I give up for now..

PR #337, opened 2020-02-11
Take event delay into account when warning about long pickup

Pchelolo:
Bug: T244758

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -411,8 +411,7 @@ class BaseExecutor {
 
         if (origEvent.delay_until && !this.disable_delayed_execution) {
             // The delay_until in the job schema is a timestamps in seconds
-            // TODO: Suppport both ISO 8601 and mediawiki date formats. Remove after transition.
-            const delayUntil = Date.parse(origEvent.delay_until) || origEvent.delay_until * 1000;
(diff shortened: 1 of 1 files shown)


lib/base_executor.js
```diff
@@ -457,7 +456,7 @@ class BaseExecutor {
         }));
 
         const metricStartTime = statDelayStartTime || new Date(origEvent.meta.dt);
-        if (Date.now() - metricStartTime > 86400000) {
+        if (Date.now() - (Date.parse(origEvent.delay_until) || metricStartTime) > 86400000) {
```
clarakosi: I'm a little confused here. When a delayed event is processed does `this.disable_delayed_execution` become true?

Pchelolo: We don't change to this.disable_delayed_execution, it's taken from a rule configuration.

Pchelolo:
> On the ticket, it says: `1 day passed between job release timestamp and execution, and not job create timestamp and execution.` What exactly do you mean by job release timestamp?

I mean the time when the delayed job is scheduled to be executed. We're holding it off until 'delay_until' and then release it. So really 'job release timestamp' means 'delay_until'..

lib/base_executor.js
clarakosi: 👍

PR #341, opened 2020-03-18
Replace /page/media references with /page/media-list

mdholloway:
Reference /page/media-list in the example Wikimedia config, not
/page/media, which will soon be removed.


Pchelolo:
Travis failures unrelated.

PR #348, opened 2020-06-12
Update dependencies and drop node 6 support.

Pchelolo:
This PR does the following:
- Require node 10
- Update all dependencies
- Bring the codebase into complience with the latest eslint rules
- Most importantly - we've been discussing that building librdkafka from sources
instead of relying on the packages hosted on apt.wikimedia.org is beneficial,
since we cut the last dependency between services using kafka. This PR makes
it build librdkafka from sources included in node-rdkafka. This guarantees we
would never have mismatching versions.
- Drop "deploy" stanza from the config - not needed anymore.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -27,3 +27,6 @@ rules:
     - off
   no-multi-spaces:
     - off
+  jsdoc/no-undefined-types:
+    - off
+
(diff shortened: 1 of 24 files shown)


nosmo:
lgtm!

akosiaris:
Mostly LGTM, but I do have a question

>This PR makes it build librdkafka from sources included in node-rdkafka. This guarantees we
would never have mismatching versions.

Why does this guarantee we won't have mismatching versions? And mismatching versions between which services?

Pchelolo:
> Why does this guarantee we won't have mismatching versions? And mismatching versions between which services?

The guarantee that node-rdkafka and librdkafka version are always matching. node-rdkafka has the sources of the correct librdkafka vendored, so it always builds the correct version.

PR #351, opened 2020-08-18
Change partition default to null

clarakosi:
Bug: T157649


Pchelolo:
Ok, let's wait for travis, but I think it's neat. One thing you can try doing to have more integration coverage is to change the number of partitions on some transcludes topics in tests/utils/test_topics (just add ` 2` in the end of line for test_dc.change-prop.transcludes.resource-change for example)

PR #352, opened 2020-11-02
base_executor: Improve logging and retry failed message commits

clarakosi:
Bug: T93049

### DIFF
--- a/lib/base_executor.js
+++ b/lib/base_executor.js
@@ -100,9 +100,29 @@ class BaseExecutor {
         );
 
         this._commitTimeout = null;
-        // In order ti filter out the pending messages faster make them offset->msg map
+        // In order to filter out the pending messages faster make them offset->msg map
         this._pendingMsgs = new Set();
         this._pendingCommits = new Map();
+        this._selfCheck = setInterval(() => {
+            for (const message of this._pendingMsgs.entries()) {
+                this._logger.log('error/commit_check', () => ({
+                    message: 'Commit failed',
+                    offset: message.offset,
+                    raw_event: message.value.toString(),
+                    description: 'Pending message is older than 1 minute'
+                }));
+            }
+            for (const commitQueue of this._pendingCommits.values()) {
+                commitQueue.forEach(message => {
+                    this._logger.log('error/commit_check', () => ({
+                        message: 'Commit failed',
+                        offset: message.offset,
+                        raw_event: message.value.toString(),
+                        description: 'Pending commit is older than 1 minute'
+                    }));
+                });
+            }
+        }, 60000);
         this._consuming = false;
         this._connected = false;
     }
@@ -322,6 +342,7 @@ class BaseExecutor {
                 }
                 return P.all(toCommit.map(message => this.consumer.commitMessageAsync(message)
                     .catch((e) => {
+                        this._pendingCommits.get(message.topic).push(message);
                         this._logger.log('error/commit', () => ({
                             message: 'Commit failed',
                             offset: message.offset,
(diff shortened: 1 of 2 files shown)


lib/base_executor.js
```diff
@@ -322,6 +342,7 @@ class BaseExecutor {
                 }
                 return P.all(toCommit.map(message => this.consumer.commitMessageAsync(message)
                     .catch((e) => {
+                        this._pendingCommits.get(message.topic).push(message);
```
Pchelolo, line 345: Please add a comment like 'put the message back to re-try commit on next pass' or something alike.

lib/base_executor.js
```diff
@@ -100,9 +100,29 @@ class BaseExecutor {
         );
 
         this._commitTimeout = null;
-        // In order ti filter out the pending messages faster make them offset->msg map
+        // In order to filter out the pending messages faster make them offset->msg map
```
Pchelolo, line 103: I believe this is not correct anymore.

lib/base_executor.js
```diff
@@ -100,9 +100,29 @@ class BaseExecutor {
...
         this._pendingMsgs = new Set();
         this._pendingCommits = new Map();
+        this._selfCheck = setInterval(() => {
+            for (const message of this._pendingMsgs.entries()) {
+                this._logger.log('error/commit_check', () => ({
+                    message: 'Commit failed',
+                    offset: message.offset,
+                    raw_event: message.value.toString(),
+                    description: 'Pending message is older than 1 minute'
```
Pchelolo, line 112: You never actually check that it's older then 1 minute do you? this will just dump everything that's currently pending out there

lib/base_executor.js
```diff
@@ -100,9 +100,29 @@ class BaseExecutor {
...
+                }));
+            }
+            for (const commitQueue of this._pendingCommits.values()) {
+                commitQueue.forEach(message => {
+                    this._logger.log('error/commit_check', () => ({
+                        message: 'Commit failed',
+                        offset: message.offset,
+                        raw_event: message.value.toString(),
+                        description: 'Pending commit is older than 1 minute'
```
Pchelolo, line 121: same.

Pchelolo (changes requested):

lib/base_executor.js
```diff
@@ -100,9 +100,32 @@ class BaseExecutor {
         );
 
         this._commitTimeout = null;
-        // In order ti filter out the pending messages faster make them offset->msg map
         this._pendingMsgs = new Set();
         this._pendingCommits = new Map();
+        this._selfCheck = setInterval(() => {
+            for (const message of this._pendingMsgs.entries()) {
+                // If pending message is older than 1 minute then log error
+                if ((Date.now() - message.timestamp) > 60000) {
```
Pchelolo, line 108: nope :) message.timestamp is when the message was sent to kafka, not when we dequeued it. I guess we need to store 
{
 message: message,
 dequeue_timestamp: Date.now()
}
in the ending messages queue

lib/base_executor.js
```diff
@@ -100,9 +100,32 @@ class BaseExecutor {
...
+                        message: 'Commit failed',
+                        offset: message.offset,
+                        description: 'Pending message is older than 1 minute'
+                    }));
+                }
+            }
+            for (const commitQueue of this._pendingCommits.values()) {
+                commitQueue.forEach(message => {
+                    // If pending commit is older than 1 minute then log error
+                    if ((Date.now() - message.timestamp) > 60000) {
```
Pchelolo, line 119: same.

lib/base_executor.js
Pchelolo: or, we can add a 'dequeue_timestamp' property to the message when we read it. That's better.

Pchelolo (approved):

